PR #10, opened 2026-03-03
add Nix development environment

winterqt:


carlsverre:
thoughts on a github action which validates all the nix stuff works so it doesn't break randomly?

carlsverre:
🥳

PR #11, opened 2026-03-03
run: build config container via -c / --config

lmb:
Every user has to provide a config image to kick off a run. So far that means writing the same dummy Dockerfile over and over.

Add a flag to the run command which builds a config container from a directory, pushes that to the registry and then uses it to start a run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,12 +57,40 @@ The `-w`/`--webhook` flag specifies which webhook to call. Common values are `ba
 
 ### Launch a test run
 
-The `-c`/`--config` flag points at a local directory containing your `docker-compose.yml`. Snouty builds a config image from that directory and pushes it to `ANTITHESIS_REPOSITORY` automatically.
+The `-c`/`--config` flag points at a local directory containing your `docker-compose.yaml`. Snouty builds a config image from that directory and pushes it to `ANTITHESIS_REPOSITORY` automatically.
 
-```
+```sh
 snouty run --webhook basic_test --config ./config \
+  --test-name "my-test" \
+  --description "nightly test run" \
+  --duration 30 \
+  --recipients "team@example.com"
+```
+
+Alternatively, pass a pre-built config image directly:
+
+```sh
+snouty run --webhook basic_test \
+  --config-image us-central1-docker.pkg.dev/proj/repo/config:latest \
+  --duration 30
+```
+
+Extra parameters (e.g. integration tokens, custom properties) can be passed with `--param`:
+
+```sh
+snouty run -w basic_test --duration 30 \
+  --param antithesis.integrations.github.token=TOKEN \
+  --param my.custom.property=value
+```
+
+### Raw API access
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -48,19 +48,23 @@ Set the following environment variables:
 export ANTITHESIS_USERNAME="your-username"
 export ANTITHESIS_PASSWORD="your-password"
 export ANTITHESIS_TENANT="your-tenant"
+export ANTITHESIS_REGISTRY="us-central1-docker.pkg.dev/your-project/your-repo"
```
lmb: Not sure whether this should be ANTITHESIS_REPOSITORY instead.

src/cli.rs
```diff
@@ -14,18 +14,33 @@ pub enum Commands {
     #[command(long_about = r#"Launch a test run
 
 Example:
-  snouty run -w basic_test \
+  snouty run -w basic_test -c ./config \
     --antithesis.test_name "my-test" \
     --antithesis.description "nightly test run" \
-    --antithesis.config_image config:latest \
     --antithesis.images app:latest \
```
carlsverre: thoughts on removing the images arg by default and just having that as an optional field? I think image inference makes it obsolete

README.md
carlsverre: https://stackoverflow.com/a/34004418/65872 seems to say that `us-central1-docker.pkg.dev` is the registry while `your-project/your-repo` is the repository.

so probably repository? or maybe just w/e we tell customers when we give them the url

carlsverre (commented):
General comment: thoughts on renaming `snouty run` to `snouty webhook` and keeping it raw, then deprecating `snouty run` (but keeping it as an alias to webhook), and then having this new more fancy version be `snouty submit`?

A bit of pain now, but leading towards splitting between fancy/raw. Sorta similar to how the aws cli has `s3` and `s3api` subcommands. one nice one and one raw one.

all suggested names up for discussion

carlsverre:
Pure rust image gen + push is very compelling, but how much does it increase the support surface area? Is it less or more likely we run into weird "it doesn't work on my machine" tickets? My main concern would be osx/windows compat.

src/cli.rs
```diff
@@ -66,3 +69,80 @@ Using Moment.from (copy from triage report):
...
+    #[arg(long)]
+    pub description: Option<String>,
+
+    /// Test duration in minutes
+    #[arg(long)]
+    pub duration: Option<u32>,
+
+    /// Container images (semicolon-delimited)
+    #[arg(long)]
+    pub images: Option<String>,
```
carlsverre: wonder if this can just be a `--param antithesis.images`? I'm curious when this param is actually needed now that we have image inference.

carlsverre (approved):
Hope you're ok with me just editing the PR rather than leaving a bunch of comments. This is a sweet diff, let's get it landed.

src/cli.rs
carlsverre: while you're at it might as well migrate debug under api.

PR #12, opened 2026-03-03
switch from thiserror to eyre

carlsverre:


carlsverre:
Shrug, I am more used to anyhow. From reading the docs eyre feels like a lot. Do you have a strong pref?

PR #15, opened 2026-03-05
add agents.md and specs workflow

carlsverre:


lmb:
@carlsverre I like the idea, but I wonder if it would be better if the specs were executable? Two examples that come to mind:

1. Something like showboat: https://simonwillison.net/2026/Feb/10/showboat-and-rodney/
2. The go scripttest package / [txtar](https://docs.rs/txtar/latest/txtar/) format: https://atlasgo.io/blog/2024/09/09/how-go-tests-go-test#2018--script_testgo

This is basically expect style tests in a small DSL, with comments interspersed and support for specifying input files. This can be executed as a unit test which gives the agent a signal to work against.

PR #58, opened 2026-03-20
Validate Command: Copy results to SNOUTY_TEMP_DIR

SeanCheatham:
- Adds support for environment variable `SNOUTY_TEMP_DIR`
  - This is not a "documented" feature; it is experimental at the moment
- This allows the snouty user to retain the Antithesis SDK output and all applied workload scripts for further analysis
- Directory structure:
  - `$SNOUTY_TEMP_DIR/`
    - `scripts`
      - `{service_name}`
        - `scriptA`
        - `scriptB`
    - `antithesis`
      - `{service_name}`
        - `sdk.jsonl`
    - `override.yml`

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -121,6 +121,25 @@ fn generate_setup_override(compose_yaml: &str, temp_dir: &Path) -> Result<PathBu
     Ok(override_path)
 }
 
+/// Create a directory at the requested path if it does not already exist.
+/// If it already exists, ensure it is a directory, and ensure there is nothing in it.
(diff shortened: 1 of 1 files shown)


src/validate.rs
```diff
@@ -603,19 +659,21 @@ async fn watch_for_setup_complete(
             return Ok(false);
         }
 
-        for path in find_jsonl_files(output_dir)? {
```
carlsverre: does this change actually tweak anything? it seems like it makes it search incrementally... is that needed?

PR #92, opened 2026-04-08
validate: replace script execution with static structure checks

lmb:
`snouty validate` was running test scripts inside containers via `compose exec` to verify them. This creates side effects and requires a live runtime just to check structure.

## Changes

- **Removed** `exec_script`, `run_test_scripts`, `shuffle`, and `diagnose_setup_in_first_scripts` from `validate.rs`
- **Added** `validate_test_scripts` — static check that scripts have valid prefixes and at least one `driver` or `anytime` script is present; no containers are exec'd
- **Moved** script scanning + type classification into a new `src/scripts.rs` module (`ScriptType`, `TestScript`, `scan_scripts`, `ScanResult`) with `not_executable` detection via filesystem permission bits
- **Updated** specs: removed `validate_timeout.txt` (no longer relevant), trimmed `validate_failures.txt` (dropped `fail_with_finally` fixture), updated `validate_setup.txt` and `help.txt` to reflect the new static-only behavior
- **Removed** `--timeout` flag from `ValidateArgs` (no longer needed without exec)


src/validate.rs
```diff
@@ -255,10 +255,9 @@ async fn validate_with_temp_dir(args: ValidateArgs, temp_dir: &Path) -> Result<(
...
         Ok(true) => {
             eprintln!("Setup-complete event detected.");
-            run_test_scripts(&*compose, &config, &scripts)
+            validate_test_scripts(&scripts)
         }
         Ok(false) => {
-            diagnose_setup_in_first_scripts(&*compose, &config, &scripts, temp_dir);
             bail!("timed out waiting for setup-complete event");
         }
         Err(e) => Err(e),
```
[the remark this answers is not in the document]
lmb: @copilot fix this by updating the comment.

carlsverre:
@lmb I totally lost track of this. Is this ready to go?

src/validate.rs
```diff
@@ -397,6 +335,13 @@ fn discover_scripts(
...
+                if !result.not_executable.is_empty() {
+                    let mut err = eyre!("scripts in service {service_name} are not executable");
+                    for command in result.not_executable {
+                        err = err.with_note(|| command);
+                    }
+                    return Err(err);
+                }
                 if result.scripts.is_empty() {
                     bail!("No scripts found in {service_name}");
                 }
```
carlsverre, line 347: IMO this shouldn't be a failure. I believe Antithesis is ok with empty test template dirs.

src/validate.rs
```diff
@@ -416,143 +361,43 @@ fn discover_scripts(
...
     );
 
     if scripts.is_empty() {
         debug!("no services contained test scripts");
         return Ok(());
     }
 
-    if drivers.is_empty() && anytime.is_empty() {
+    if drivers == 0 && anytime == 0 {
         bail!("test scripts found but no driver or anytime scripts");
```
carlsverre: seems like this error may also be excessive.

carlsverre (commented):
if you can get confirmation on the empty template thing that would be great. but assuming I'm right and antithesis is ok with it, I think we should let it go through. maybe raise it as a warning, but not an error

lmb:
Had a quick check and can't find where I got that from. Made the changes you suggestd.

PR #105, opened 2026-05-15
v0 API support

lmb:
Generate the API client from the OpenAPI spec at build time instead of maintaining bindings by hand. Building on that:

- add a `snouty runs` subcommand for listing and inspecting runs
- add an HTTP cache layer that honors Cache-Control and ETag so repeated reads can be served from disk or revalidated cheaply
- add a mock API harness so end-to-end tests can run offline, with `SNOUTY_STAGING=1` switching the same specs to hit a real backend

Existing CLI commands and their spec snapshots are updated to align with the generated types.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
         run: cargo fmt --check
 
       - name: Build
-        run: cargo build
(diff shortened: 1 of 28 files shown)


lmb:
@claude review this

PR #112, opened 2026-05-26
Port fault annotation from skills

### LINKED ISSUE #96: logs: Port fault injector analysis from skills
Our skills have a python script which annotates fault_injector logs to make it clearer which faults are active at a given time. See https://github.com/antithesishq/antithesis-skills/blob/main/antithesis-triage/assets/process-logs.py

We should port this to the logs subcommand:

```
snouty runs logs --annotate-faults ...
snouty --json runs logs --annotate-faults ... # outputs NDJSON
```

TBD:
- Name of the flag
- Default to on or off? (Probably `on`)

jeskew:
Resolves #96 

This PR is meant to replicate the effect of running [process-logs.py](https://github.com/antithesishq/antithesis-skills/blob/main/antithesis-triage/assets/process-logs.py) over a log file. Annotations are applied over the log stream as new entries come in.

I'm pretty new to rust, so let me know if anything should be updated to match the expected style or be more idiomatic.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -352,7 +352,7 @@ pub enum RunsCommands {
         run_id: String,
 
         /// Whether to annotate faults that appear in individual log lines
-        #[arg(long, default_value_t = true)]
+        #[arg(long, default_value_t = false, requires = "json" )]
         annotate_faults: bool,
     },
 
@@ -377,7 +377,7 @@ pub enum RunsCommands {
         begin_input_hash: Option<String>,
 
         /// Whether to annotate faults that appear in individual log lines
-        #[arg(long, default_value_t = true)]
+        #[arg(long, default_value_t = false, requires = "json" )]
         annotate_faults: bool,
     },
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -389,10 +389,6 @@ pub enum RunsCommands {
         /// Search query
         #[arg(required = true, num_args = 1..)]
         query: Vec<String>,
-
-        /// Whether to annotate faults that appear in individual log lines
-        #[arg(long, default_value_t = true)]
-        annotate_faults: bool,
     },
 }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -350,6 +350,10 @@ pub enum RunsCommands {
     BuildLogs {
         /// Run ID
         run_id: String,
+
+        /// Whether to annotate faults that appear in individual log lines
+        #[arg(long, default_value_t = true)]
+        annotate_faults: bool,
     },
 
     /// Stream moment logs for a run
@@ -371,6 +375,10 @@ pub enum RunsCommands {
         /// Start streaming from this input hash (optimization; must be paired with --begin-vtime)
         #[arg(long, allow_hyphen_values = true, requires = "begin_vtime")]
         begin_input_hash: Option<String>,
+
+        /// Whether to annotate faults that appear in individual log lines
+        #[arg(long, default_value_t = true)]
+        annotate_faults: bool,
     },
 
     /// Search events in a run
@@ -381,6 +389,10 @@ pub enum RunsCommands {
         /// Search query
         #[arg(required = true, num_args = 1..)]
         query: Vec<String>,
+
+        /// Whether to annotate faults that appear in individual log lines
+        #[arg(long, default_value_t = true)]
+        annotate_faults: bool,
     },
 }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -350,10 +350,6 @@ pub enum RunsCommands {
     BuildLogs {
         /// Run ID
         run_id: String,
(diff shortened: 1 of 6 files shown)


src/cli.rs
```diff
@@ -350,6 +350,10 @@ pub enum RunsCommands {
     BuildLogs {
         /// Run ID
         run_id: String,
+
+        /// Whether to annotate faults that appear in individual log lines
+        #[arg(long, default_value_t = true)]
+        annotate_faults: bool,
```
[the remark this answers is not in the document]
jeskew: Defaulting to off unless a reviewer requests otherwise. This should be a non-breaking change for users even if it defaults to `on`, but the point about potentially introducing instability for existing users is taken.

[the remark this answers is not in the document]
jeskew: Added a spec test

src/runs.rs
```diff
@@ -222,21 +228,29 @@ fn print_run_detail(run: &RunDetail) {
...
+        } else {
+            stream_ndjson_lines(stream, NoOpTransformer {}, |line| {
+                writeln!(stdout, "{line}")?;
+                Ok(())
+            })
+            .await
+        }
     } else {
-        stream_ndjson_lines(stream, |line| {
+        stream_ndjson_lines(stream, NoOpTransformer {}, |line| {
```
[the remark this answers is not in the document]
jeskew: Made `--annotate-faults` require `--json`. If there's a good display option for the non-NDJSON output format, this can be removed.

carlsverre:
@claude review

src/cli.rs
```diff
@@ -350,6 +350,10 @@ pub enum RunsCommands {
     BuildLogs {
         /// Run ID
         run_id: String,
+
+        /// Whether to annotate faults that appear in individual log lines
+        #[arg(long, requires = "json")]
+        annotate_faults: bool,
```
carlsverre: @jeskew wdyt about making this the default and having an opt-out flag instead? I see the discussion between Claude and you, but IMO, this generally improves the logs. The only reason not to do it is likely if it is somehow broken due to the server changing the log format or something similar.

carlsverre: (same in the other command)

src/runs.rs
```diff
@@ -379,6 +559,163 @@ struct RenderedEventEntry {
...
+
+    fn is_network_fault(&self) -> bool {
+        match self {
+            Self::Network {
+                name: _,
+                start_vtime: _,
+                end_vtime: _,
+            } => true,
+            _ => false,
+        }
```
carlsverre: nit, this pattern can be replaced with the `matches!` macro along with `..`:
```rust
matches!(self, Self::Network { .. })
```

carlsverre: same elsewhere

src/runs.rs
```diff
@@ -1304,4 +1641,415 @@ mod tests {
             r"a\x01b\x0Bc\x7Fd	e"
         );
     }
+    #[test]
```
carlsverre, line 1699: did you copy all of the tests from process_logs.py?
https://github.com/antithesishq/antithesis-skills/blob/f8372481aa538b6484706ccb683b9f3113a633d4/antithesis-triage/assets/process-logs.py#L275

carlsverre (commented):
I think we need to find a way to test this against actual log files. Thoughts on adding some snapshot tests? The tricky part is finding some example logs we can share in an open-source repo.

In general, I'm pretty happy with this PR. We can always iterate on it in main. LMK if you're good to go or want to tweak any of the last comments. I can also have Claude fix up the remaining items in a followup PR.

src/cli.rs
jeskew: I agree that an opt-out is preferable. I think this means the flag has to use negative language (i.e., `--disable-fault-annotation` or `--no-annotate-faults`) because I couldn't figure out how to make Clap accept an argument for a boolean flag (like `--annotate-faults=false`).

src/runs.rs
jeskew: I had Claude port the rest

src/runs.rs
```diff
@@ -379,6 +551,223 @@ struct RenderedEventEntry {
...
+                        if let Some(updated) = merge_fault_windows(entry.get(), window) {
+                            entry.insert(updated);
+                            return true;
+                        }
+
+                        false
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jeskew: This was an intentional divergence from the python source. It's odd that network fault window annotations report when the fault began, but node faults report when the window was last renewed. Aligning them provides more internal consistency. 

Clock faults use the latest timestamp because the accumulation of offsets is only accurate as of the last clock fault injection. Node faults, on the other hand, are like network faults in that they are either on or off, so when the fault window was last extended is just not very useful information.

PR #116, opened 2026-06-01
feat: load env variables from .env file if exists

m-kus:
- snouty looks for a .env file in the current directory, walking up to parent
  directories if none is found
- reads all config from it: API key, username, password, tenant, repo
- real shell env vars take precedence over .env — no override, the shell wins
- if there's no .env, snouty works exactly as before — nothing breaks
- adds the dotenvy crate, loaded before anything reads the environment
- .env is added to .gitignore so secrets don't leak into git
- a Claude settings rule denies reading .env files, so the agent can't see secrets

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "deny": [
+      "Read(./.env)",
+      "Read(./.env.*)"
+    ]
+  }
+}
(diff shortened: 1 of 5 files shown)


carlsverre:
Hey! Thanks for the PR. We are about to add an official config file to snouty which will support this use case as well as integrate with our upcoming oauth feature for better auth. Hope to have the config change out asap!

m-kus:
Cool, thanks!

Having interactive `snouty auth login` (a-la gcloud) would be perfect!

PR #123, opened 2026-06-04
Improve human-facing runs output

carlsverre:
This branch reworks the `runs` command family to produce output that reads well for humans, and adds the tooling to keep that output honest.

The bulk of the change is in `src/runs.rs`, which now renders run listings, properties, and events with human-oriented formatting rather than raw API dumps. The default `runs` table drops the DESCRIPTION column (it never fit usefully beside the full run id — `runs list --long` shows it instead), and the properties table is reordered to STATUS/EXAMPLES/NAME with the group folded into the name as `group ▸ name`. Supporting pieces: the CLI accepts negative vtime bounds for runs events, and properties-endpoint 404s are now explained — a missing run is distinguished from a real run that simply has no triage report yet, instead of leaking a bare "404 Not Found".

To exercise and review this output, `scripts/gen-gallery.py` is a rewritten gallery generator (replacing the old build.rs-driven path) that captures live-API output into gallery stories, with per-run timeout recovery and a bumped client timeout. The new `satisfaction-review` skill drives an LLM review over those captured stories to judge how good the CLI output actually is.

The `specs/runs.txt` assertions are updated to match the new table layouts, and the gallery script clears two ruff findings.


carlsverre:
@claude review

PR #126, opened 2026-06-04
fix: always build config image for amd64

carlsverre:
Antithesis only supports x86-64 (amd64) container images, but `build_and_push_config_image` was building with no explicit platform. On an arm64 host that meant the auto-generated config image got stamped `arm64` and rejected by the platform.

This passes `--platform linux/amd64` for the config-image build. Because the config image is `FROM scratch` + `COPY . /` with no `RUN` steps, there's no QEMU emulation involved, so it builds on any standard Docker/Podman install without needing multi-platform support enabled — and it adds no new build requirement, since this is the trivial no-emulation cross-arch case.

### DIFF
--- a/src/container.rs
+++ b/src/container.rs
@@ -237,9 +237,12 @@ pub trait ContainerRuntime: Send + Sync {
     /// Callers are responsible for validating the directory beforehand
     /// (typically via [`crate::config::detect_config`]). Returns the pinned
     /// image reference.
+    ///
+    /// The image is always built for `linux/amd64` (x86-64) because Antithesis
+    /// does not support arm64, and the host may well be an arm machine.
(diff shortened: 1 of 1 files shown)


lmb:
Arguably the config image doesn't really have a platform? I'm guessing that building on arm64 defaults to arm64, but the real fix is to somehow explicitly say "platform independent".

carlsverre:
yea I investigated that (via ai) and it seems like there is no good way to make an image platform agnostic. it's possible to build a multi-platform image, however that may require a multi-plat compatible build system (i.e. buildx) which is not always available. since the config image is only used by our platform and afaik not useful to the user (basically an implementation detail of launching a run), I figured it would be fine to just use what we expect directly.

PR #127, opened 2026-06-08
feat!: require docker-compose and digest-pin compose images

### LINKED ISSUE #98: drop `podman-compose` support?
snouty prefers `podman compose` over `docker compose`. `podman compose` defaults to invoking `podman-compose` (not the same thing!) but optionally uses `docker-compose` if the user has configured it correctly. `docker-compose` can then talk to the podman image store and do its thing.

podman-compose is pretty janky. people are confused by the fact that we prefer podman over docker, which leads to using podman-compose instead of docker-compose.

the original sin here might be that we're trying to auto detect what engine to use. or it might be that we tried to support podman-compose.

### LINKED ISSUE #113: stop using antithesis.images and rely on rewrite+pull semantics
currently snouty launch rewrites the docker compose file to pin all images and then also submits the referenced images via the images param on the launch webhook endpoint. this seems to no longer be needed, as antithesis will automatically pull referenced images.

this needs to be tested and verified, but we should be able to drop the images param

carlsverre:
snouty now always drives Docker Compose v2 through the `docker-compose` binary rather than `<runtime> compose`, dropping podman-compose support. podman is still supported as the image runtime for build/push; when it's the runtime, docker-compose is pointed at podman's API socket via `DOCKER_HOST` (an explicit `DOCKER_HOST` already in the environment is respected). This removes the podman-compose backend, whose jank and ambiguous `podman compose` dispatch were the original problem.

For compose configs, `snouty launch` now rewrites `docker-compose.yaml` with digest-pinned images and bakes that into the config image, so a re-run pulls exactly the images resolved at launch instead of re-resolving floating tags. Images are resolved registry-first:

- a `build:` stanza pushes the local build to the tenant registry and pins it;
- any other image is pinned from the registry it names — third-party refs like `redis:7` or `nginx` pin from Docker Hub and are never pushed, even when also present locally;
- an image present only locally (built or loaded out of band, no `build:` stanza) is pushed to the tenant registry and pinned;
- an image that is neither present locally nor pullable from a registry is a hard error, surfaced in seconds locally instead of a failed image pull minutes into a run.

Digests are read via the container runtime itself (pull + the image's repo digest), so resolution works identically on docker and podman — `docker-compose config --resolve-image-digests` was avoided because it relies on a daemon distribution endpoint podman's API doesn't implement. Pinned references keep their tag (`name:tag@sha256:...`) as provenance. snouty no longer sets `antithesis.images` — the pinned compose file is the source of truth.

Dropping the second compose backend also removed the flavor detection, the podman-compose `ps`/`logs` workarounds, and the `Compose` trait, collapsing it to a single `DockerCompose` type.

CI was updated to install the `docker-compose` binary on each runner and to drop the obsolete `podman-compose` matrix combinations; `shell.nix` gained `docker-compose` for the nix shell.

On macOS, podman recomputes its machine API socket path from the current `TMPDIR` on every invocation (falling back to `/tmp` when unset), while the socket was bound under the `TMPDIR` in effect at `podman machine start`. The spec-test harness scrubs the environment, which dropped `TMPDIR` and pointed docker-compose at a socket path that didn't exist on the macOS runners. The harness now forwards `TMPDIR`, and `engine_docker_host` probes sibling socket locations when the reported path doesn't exist, so users running snouty from an environment without `TMPDIR` (launchd services, some CI) still reach the machine socket.

Upgrading requires the `docker-compose` binary (Docker Compose v2) to be installed. `snouty doctor` now checks for it, and errors are clear when it's missing.

Closes #98
Closes #113

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,16 +70,10 @@ jobs:
         include:
           - os: ubuntu-latest
             runtime: docker
-            compose: docker-compose
           - os: ubuntu-latest
             runtime: podman
-            compose: podman-compose
-          - os: ubuntu-latest
-            runtime: podman
-            compose: docker-compose
           - os: macos-15-intel
             runtime: podman
-            compose: podman-compose
     env:
       SNOUTY_TEST_RUNTIME: ${{ matrix.runtime }}
     steps:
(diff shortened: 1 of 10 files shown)


carlsverre:
(this is currently fully vibed, will do a human review tomorrow)

carlsverre:
@claude review

PR #128, opened 2026-06-11
File-based Snouty configuration

### LINKED ISSUE #120: Snouty config from files + precedence
Somewhat related to #107, Snouty may be outgrowing its environment-variables-only configuration strategy. Supporting OAuth will mean persisting some amount of state across invocations, and there may be auth-related configuration that would be onerous to manage on the environment or otherwise explicitly supply on each call to `snouty`.

I would propose that we follow the model used by tools like cloud provider CLIs and Claude code where we merge configuration from a few sources with a well-defined precedence order as follows:

# Environment variables

For both backwards compatibility and parity with other CLI tools, environment variables should take precedence over anything read from disk. 

We should continue to support the following environment variables:

<dl>
<dt>ANTITHESIS_TENANT</dt>
<dd>The Antithesis tenant (i.e., subdomain) you want to interact with.</dd>
<dt>ANTITHESIS_REPOSITORY</dt>
<dd>The container repository you use to push and pull images</dd>
<dt>ANTITHESIS_BASE_URL</dt>
<dd>The base URL to use to talk to Antithesis. Defaults to <code>https://&lt;tenant&gt;.antithesis.com</code>.</dd>
<dt>ANTITHESIS_DOCS_DB_PATH</dt>
<dd>Where to find the docs database.</dd>
<dt>ANTITHESIS_DOCS_URL</dt>
<dd>The base URL to use for displaying docs. Defaults to <code>https://antithesis.com/docs</code>.</dd>
<dt>SNOUTY_CONTAINER_ENGINE</dt>
<dd>Which local container engine to use, either <code>podman</code> or <code>docker</code>. Defaults to <code>podman</code>.</dd>
<dt>SNOUTY_TEST_RUNTIME</dt>
<dd>Which local container engine to use, either <code>podman</code> or <code>docker</code>. Defaults to <code>podman</code>.</dd>
<dt>SNOUTY_TEST_CACHE_DIR</dt>
<dd>The cache directory to use</dd>
<dt>SNOUTY_TEMP_DIR</dt>
<dd>The temp directory to use</dd>
<dt>ANTITHESIS_API_KEY</dt>
<dd>The credential to use for API key authentication</dd>
<dt>ANTITHESIS_USERNAME</dt>
<dd>The username to use for basic authentication</dd>
<dt>ANTITHESIS_PASSWORD</dt>
<dd>The password to use for basic authentication</dd>
</dl>

This list could probably be cleaned up a bit -- are `SNOUTY_CONTAINER_ENGINE` and `SNOUTY_TEST_RUNTIME` really distinct configuration options? -- but that doesn't need to happen immediately.

We should also add a new, optional `ANTITHESIS_PROFILE` environment variable, whose role is described in the next session.

# Configuration files

## Format

Snouty should support TOML configuration files with the following top-level keys:

<dl>
<dt>tenant</dt>
<dd>The Antithesis tenant (i.e., subdomain) you want to interact with.</dd>
<dt>repository</dt>
<dd>The container repository you use to push and pull images</dd>
<dt>base_url</dt>
<dd>The base URL to use to talk to Antithesis. Defaults to <code>https://<tenant>.antithesis.com</code>.</dd>
<dt>cache_dir</dt>
<dd>The cache directory to use. Defaults to something OS-specific (e.g., `$XDG_CACHE_HOME/snouty`).</dd>
</dl>

Additionally, Snouty should support a top-level `profiles` key whose value is an object, whose values may have the same keys as would be supported at the top level of the config file (with the exception of `profiles`). 

### Example

```toml
tenant = "your-tenant"
repository = "us-central1-docker.pkg.dev/your-project/your-repo"

profiles.bar.base_url = "https://cheerio.antithesis.co.uk"

[profiles.foo]
repository = "us-central1-docker.pkg.dev/your-project/foo-repo"
```

### global

Snouty should look for a global configuration file named `settings.toml` at a path that varies by OS. E.g., `$XDG_CONFIG_HOME/snouty/settings.toml` on Linux or `$HOME/Library/Application Support/com.Antithesis.Snouty/settings.toml`, with the exact directory determined by [the `directories-next` crate](https://crates.io/crates/directories-next).

### Project-specific

For Snouty command that accept a configuration directory (e.g., `snouty launch -c ~/path/to/project` or `snouty validate ./projects/foo`), we should also look for a configuration file at `<config dir>/snoutyconfig.toml`. (Very open to alternative suggestions on the filename here, though I do think it's important to include `snouty` somehow to distinguish it from other, project-specific files.)

Commands that don't currently accept a configuration directory **should not start doing so**, as those commands are designed to interact only with the Antithesis API and not the filesystem. Customers who need to use different configuration for different invocations of, say, `snouty runs logs ...` should use profiles rather than pointing at different files. The rationale for this is that CLI tools typically only want to support project-specific configuration that is colocated with other project files so that it can be committed alongside them.

# Merge behavior

Configuration need not all come from the same source, as there are cases where you have defined a configuration file but want to override a specific value for a single invocation. Snouty should look up each configuration value with a defined precedence ordering and without regard to where other configuration values were sourced. For example, fetching the tenant would look something like:

1. Check for `ANTITHESIS_TENANT`. If found, **return its value**.
2. Resolve active profile name from `ANTITHESIS_PROFILE`. Bind to `$profile` if present
3. If command has a config dir, check for `<config dir>/snoutyconfig.toml`. Bind to `$projectConfig` if present.
4. Check for the global config file. Bind to `$globalConfig` if present.
5. If `$projectConfig.profiles.$profile.tenant` is present, **return its value**.
6. If `$globalConfig.profiles.$profile.tenant` is present, **return its value**.
7. If `$projectConfig.tenant` is present, **return its value**.
8. If `$globalConfig.tenant` is present, **return its value**.
9. Raise tenant not found error.

# Potential points of contention

1. Profile-specific global configuration trumps non-profile-specific project configuration. I think it would be equally defensible to say that `$projectConfig.tenant` takes precedence over `$globalConfig.profiles.$profile.tenant`, but we should pick a precedence ordering and then use it consistently.
2. API keys and basic auth credentials should only come environment variables, not config files. OAuth-related configuration that is not sensitive can be stored in config files when that feature is ready, but credentials should not be.

From the tools I have worked on in the past, profiles typically are used for tools that don't interact with files on disk, and project-specific config is used for tools that you point at a specific directory. Snouty has some commands that fall into each category, so I think we would benefit from mixing those models, but definitely open to other opinions on this one!

I feel pretty strongly about the second point, though. We should aim to keep Snouty config files as something you could commit to source control if you were so inclined.

jeskew:
Closes #120 

This PR adds support for reading configuration from two locations on disk in addition to the currently supported environment variables. All access to configuration is routed through a `SnoutyConfig` value that will try to source the tenant and container repository from the following locations on disk if the `ANTITHESIS_TENANT` or `ANTITHESIS_REPOSITORY` environment variables are not set:

- `<current working directory>/.snouty.toml`
- `$XDG_CONFIG_HOME/snouty/settings.toml`

Both files have the same format, which looks like:

```toml
tenant = "your-tenant"
repository = "us-central1-docker.pkg.dev/your-project/your-repo"

profile.bar.tenant = "another-tenant"

[profile.foo]
repository = "us-central1-docker.pkg.dev/your-project/foo-repo"
```

Snouty will use the top-level `tenant`/`repository` values unless a profile is supplied via the `ANTITHESIS_PROFILE` environment variable or via a `--profile` arg (the arg takes precedence over the environment variable when both are set). When we add OAuth support to Snouty, any requisite configuration will come from these files. API keys and basic auth are **not** supported via the config files, as those values are sensitive and thus shouldn't appear in a plaintext config file.

One thing I want to add that is not included in this PR is a `snouty init` command that will prompt the user for values and then write out config files for them. This PR was just feeling a bit large without it added, so I thought it made sense to split it out.


jeskew:
@claude review

carlsverre:
main review:

1. switch to `.snouty.toml` for project local config
2. Let's remove directories-next in place of just manually resolving XDG config dir
3. rebase

src/settings.rs
```diff
@@ -0,0 +1,242 @@
...
+/// (project file before global file).
+fn resolve_value(
+    key: &str,
+    environment_variable_name: &str,
+    profile: Option<&str>,
+    project: Option<&Table>,
+    global: Option<&Table>,
+) -> Option<String> {
+    // The environment variable always has highest precedence.
+    if let Ok(value) = env::var(environment_variable_name) {
```
carlsverre: **Correctness — empty env var masks a valid settings-file value.** `env::var` returns `Ok("")` for an exported-but-empty var, and this returns it as `Some("")` at the highest precedence. So with `tenant = "acme"` in `.snouty.toml` but `ANTITHESIS_TENANT=` exported empty (common in CI / wrapper shells), the file value is silently ignored and `resolve_base_url()` builds `https://.antithesis.com`. An empty `ANTITHESIS_BASE_URL` is worse — `normalize_base_url("")` returns `""`, an empty client base URL.

The `--profile`/`ANTITHESIS_PROFILE` path already filters this (`.filter(|p| !p.is_empty())`); the value env vars should be treated the same way.

src/settings.rs
```diff
@@ -0,0 +1,242 @@
...
+        Err(err) => {
+            Err(err).wrap_err_with(|| format!("failed to read settings file {}", path.display()))
+        }
+    }
+}
+
+/// Resolve a single value with the precedence: environment variable, then the
+/// active profile (project file before global file), then top-level defaults
+/// (project file before global file).
+fn resolve_value(
```
carlsverre, line 276: **Correctness — a selected-but-undefined profile silently falls back to defaults.** `snouty --profile staging launch …` where `[profile.staging]` is absent (typo, or removed from the file) resolves every key from the top-level defaults with no error or warning. For a launch tool that's a real safety footgun — you think you targeted staging and you hit the default (likely prod) tenant. Consider erroring (or at least warning) when a profile is explicitly requested but defined in neither file.

src/settings.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+fn profile_value(table: &Table, profile: &str, key: &str) -> Option<String> {
+    table
+        .get(PROFILE_KEY)?
+        .as_table()?
+        .get(profile)?
+        .as_table()?
+        .get(key)?
+        .as_str()
```
carlsverre: **Correctness — mistyped (non-string) TOML values are silently dropped.** This and `default_value` (line 241) end in `.as_str()`, which is `None` for `tenant = 123` (forgot quotes) or an accidental `[tenant]` table. The value isn't just ignored — a present-but-mistyped *higher-precedence* value falls through to a lower-precedence one (a precedence violation), and the user gets the generic "No tenant found" with no hint the key was present but wrong-typed.

src/settings.rs
```diff
@@ -0,0 +1,242 @@
...
+        let profile = env::var(ANTITHESIS_PROFILE_ENV_VAR_NAME)
+            .ok()
+            .filter(|p| !p.is_empty())
+            .or_else(|| profile.filter(|p| !p.is_empty()));
+
+        // Settings file location: env var, then `--settings`, then
+        // `./.snouty.toml` (CWD). Read with `var` like every other env var
+        // here — settings values are UTF-8 (they can come from a TOML file),
+        // so the path is too.
+        let settings_path = env::var(SNOUTY_SETTINGS_PATH_VAR_NAME)
```
carlsverre: **Correctness — empty `SNOUTY_SETTINGS_PATH` hard-errors every settings-reading command.** `env::var(...).ok().map(PathBuf::from)` turns `SNOUTY_SETTINGS_PATH=` into `Some(PathBuf::from(""))`, so `settings_path_is_explicit` becomes `true` and `load_settings_file("", required=true)` returns `settings file not found:` — `doctor`/`runs`/`launch`/`validate` all abort even though `./.snouty.toml` exists. Same empty-as-unset gap as the value env vars; filter it the way the profile is filtered just above.

src/settings.rs
```diff
@@ -0,0 +1,242 @@
...
+        "No {setting_key} found: set the {environment_variable_name} environment variable, \
+         or add `{setting_key} = \"...\"` to your snouty settings file (.snouty.toml)"
+    )
+}
+
+/// `$XDG_CONFIG_HOME`/`$XDG_CACHE_HOME` if set, else `$HOME/<home_subdir>`,
+/// then joined with `snouty`. `None` only when neither the XDG var nor `HOME`
+/// is set (e.g. Windows).
+fn xdg_snouty_dir(xdg_var: &str, home_subdir: &str) -> Option<PathBuf> {
+    let base_dir = if let Ok(xdg_dir) = env::var(xdg_var) {
```
carlsverre: **Correctness — empty `XDG_CACHE_HOME`/`XDG_CONFIG_HOME` yields a relative `./snouty` path.** `if let Ok(xdg_dir) = env::var(xdg_var)` accepts `Ok("")`, so `PathBuf::from("").join("snouty")` is the *relative* path `snouty`. The XDG spec says an empty value must be treated as unset (fall back to `$HOME`). Result: the docs DB and API cache get written under the CWD and effectively move per-directory.

src/api.rs
```diff
@@ -201,103 +204,46 @@ impl Auth {
...
-    pub fn new(config: Config) -> Result<Self> {
-        let base_url = format!("https://{}.antithesis.com", config.tenant);
-        debug!("using default base URL: {}", base_url);
-        Self::with_base_url(config, base_url)
-    }
-
-    pub fn from_env(verbose: bool) -> Result<Self> {
-        Self::from_config(Config::from_env()?, verbose)
+    pub fn new(settings: &Settings, verbose: bool) -> Result<Self> {
+        let cache_dir = crate::settings::cache_dir();
```
carlsverre: **Behavioral change — API response cache moved from `$XDG_RUNTIME_DIR` to `$XDG_CACHE_HOME`.** On `main` the run-scoped read cache lives in `$XDG_RUNTIME_DIR/snouty/api-cache-v1` (tmpfs, 0700, cleared at logout) and is *disabled* when `XDG_RUNTIME_DIR` is unset. With `crate::settings::cache_dir()` it now lives under `~/.cache/snouty/...`, which means (a) caching turns *on* for the first time on macOS / cron / non-login-ssh hosts, and (b) cached private-tenant responses persist across reboots. Freshness is still server-controlled (`CacheMode::Default`), so it's not a stale-data bug, but the footprint/lifetime change is unannounced and the non-Linux path has no test coverage (the api_cache tests are `#[cfg(target_os = "linux")]`).

src/main.rs
```diff
@@ -82,35 +83,61 @@ async fn main() {
...
             Ok(())
         }
         Commands::Update(args) => cmd_update(args),
         Commands::Docs { offline, command } => docs::cmd_docs(command, offline, json).await,
+
+        // Everything else shares one settings context, resolved once up front
+        // from --settings/SNOUTY_SETTINGS_PATH (default ./.snouty.toml) and the
+        // global settings.toml.
+        command => {
+            let settings = Settings::resolve(settings_path, profile)?;
```
carlsverre: **Tradeoff to confirm — an unrelated/corrupt `./.snouty.toml` in the CWD blocks fully env-configured commands.** Resolution always parses the default project file here, and a file that exists but doesn't parse is a hard error (intentional per the PR). Because the project file is `./.snouty.toml` in the *current working directory*, a malformed file dropped by some other project/tool blocks `snouty runs list` even when `ANTITHESIS_*` env vars fully configure snouty. Worth confirming you want a hard error here vs. warn-and-continue when the environment already satisfies everything.

carlsverre (commented):
## Code review — file-based settings (#128, including the merged #141 refinements)

Reviewed the full combined state against `main`. The refactor is solid: collapsing the lazy `SnoutyConfig`/`OnceCell`/`RefCell`/`SharedReport` machinery into one eagerly-resolved `Settings` threaded by reference genuinely fixes the launch inconsistency (tenant, repository, and base_url now resolve from one instance), and marking the auth header sensitive plus the env whitelist in tests are nice tightenings.

The inline comments below are mostly edge-case correctness in the new resolution layer plus cleanup — none block the architecture. The two I'd most want fixed before merge are the **empty-env-var masking** and the **silent fallback when a selected profile doesn't exist**: both can silently send requests to the wrong tenant/host from plausible real-world configs. A common root cause across several findings: `resolve_value` and the XDG helpers treat an exported-but-empty env var as a real value, whereas the `--profile`/`ANTITHESIS_PROFILE` path deliberately normalizes empty → unset.

### Also — stale docs (couldn't anchor inline; not in the diff)
- `specs/runs.txt:10-12` still documents the API cache at `$XDG_RUNTIME_DIR/snouty/api-cache-v1`; it now lives under `$XDG_CACHE_HOME` (see the api_cache.rs comment). That block also says caching "must not change" — the location did.
- `tests/fixtures/README.md:25-28` still tells readers to export `ANTITHESIS_DOCS_DB_PATH`, which this PR removes entirely.

src/settings.rs
```diff
@@ -0,0 +1,242 @@
+//! Snouty's settings, read from a project settings file (`.snouty.toml`, in the
```
carlsverre: **Convention — the new `settings` module has no unit tests.** `AGENTS.md` › Tests states: *"Internal functions must be accompanied by unit tests."* `resolve_value`, `profile_value`, `default_value`, `load_settings_file`, and `xdg_snouty_dir` have none (the only `#[cfg(test)]` item is the `for_test_base_url` constructor), while nearly every other `src/*.rs` carries a `mod tests`. Resolution is covered only indirectly by `specs/settings.txt` — which is why the empty-string, non-string, missing-profile, and level-3-vs-4 precedence cases flagged elsewhere in this review have no guardrail.

src/container.rs
```diff
@@ -1114,13 +1115,15 @@ pub(crate) fn is_podman_in_disguise(cmd: &str) -> bool {
 /// The result is cached so detection only runs once.
 ///
 /// Set `SNOUTY_CONTAINER_ENGINE=podman` or `=docker` to force a specific runtime.
-pub fn runtime() -> Result<&'static dyn ContainerRuntime> {
+pub fn runtime(settings: &Settings) -> Result<&'static dyn ContainerRuntime> {
```
carlsverre: **Altitude — `runtime(&Settings)` ignores every caller's settings after the first.** The result is memoized in a process-wide `OnceLock`, so only the first call's `settings.container_engine()` ever matters; the signature advertises per-call configurability the cache can't honor. Safe today (one `Settings` per process) and the comment notes it, but it's a latent trap. Cleaner: resolve the engine once where settings first exist and pass a `&dyn ContainerRuntime` down, or drop the parameter and let `runtime()` read the single override itself.

src/main.rs
```diff
@@ -82,35 +83,61 @@ async fn main() {
...
             println!("snouty {}", env!("CARGO_PKG_VERSION"));
             Ok(())
         }
         Commands::Update(args) => cmd_update(args),
         Commands::Docs { offline, command } => docs::cmd_docs(command, offline, json).await,
+
+        // Everything else shares one settings context, resolved once up front
+        // from --settings/SNOUTY_SETTINGS_PATH (default ./.snouty.toml) and the
+        // global settings.toml.
+        command => {
```
carlsverre: **Simplification — double-match with an `unreachable!()` arm.** The settings-free commands are listed in the outer match and again in the inner `unreachable!("handled before settings resolution")` arm (line 137) — the same variants classified in two hand-maintained places (a third copy is `json_unaware_command_name`). Adding/moving a command can land it in the wrong branch or trip the panic instead of failing to compile. Consider expressing "needs settings" on the command type (e.g. `fn needs_settings(&Commands) -> bool`, exhaustive match), resolving once, then a single flat match.

src/container.rs
```diff
@@ -1114,13 +1115,15 @@ pub(crate) fn is_podman_in_disguise(cmd: &str) -> bool {
...
     static INSTANCE: OnceLock<Result<Box<dyn ContainerRuntime>, String>> = OnceLock::new();
 
     INSTANCE
         .get_or_init(|| {
-            if let Ok(engine) = std::env::var("SNOUTY_CONTAINER_ENGINE") {
-                return match engine.as_str() {
+            // Detection is cached process-wide on first call, so the engine
+            // override is read from the settings passed to that first call.
+            if let Some(engine) = settings.container_engine() {
```
carlsverre: **Reuse — env-var names duplicated; this branch's error literal is now orphaned.** `settings.rs` centralizes `CONTAINER_ENGINE_VAR_NAME = "SNOUTY_CONTAINER_ENGINE"` (and the `ANTITHESIS_*` names) as `pub const`, and `settings` is now the only reader of that env var — but the error string at line 1130 and the doc-comment at 1117 still hardcode `"SNOUTY_CONTAINER_ENGINE"`, and `api.rs` still hardcodes the auth var names. Since `container.rs` no longer reads the var, its literal can silently drift from the one that actually controls behavior. Reference the constants.

src/api_cache.rs
```diff
@@ -29,23 +30,25 @@ pub(crate) fn build_cached_client_at(
     ClientBuilder::new(client).with(cache).build()
 }
 
-fn cache_dir_from_runtime_dir(runtime_dir: Option<std::ffi::OsString>) -> Option<PathBuf> {
-    runtime_dir.map(|dir| PathBuf::from(dir).join("snouty").join("api-cache-v1"))
+fn cache_dir_from_runtime_dir(runtime_dir: Option<&Path>) -> Option<PathBuf> {
```
carlsverre: **Cleanup — `cache_dir_from_runtime_dir` is now misnamed and trivial.** After the refactor it's just `runtime_dir.map(|dir| dir.join("api-cache-v1"))` — the `XDG_RUNTIME_DIR` lookup and the `snouty` segment moved out. The name and the two test names (`cache_dir_uses_xdg_runtime_dir`, `cache_dir_is_absent_without_xdg_runtime_dir`) still reference a runtime dir that's no longer involved, and the tests now assert little more than `Option::map` + a literal join. Inline it into `build_cached_client` or rename.

src/api.rs
```diff
@@ -201,103 +204,46 @@ impl Auth {
...
-    #[cfg(test)]
-    fn with_base_url_and_cache_dir(
-        config: Config,
-        base_url: impl Into<String>,
-        cache_dir: std::path::PathBuf,
-    ) -> Result<Self> {
-        Self::build(config, base_url, move |client| {
-            Some(api_cache::build_cached_client_at(client, cache_dir))
-        })
+    pub fn new_requiring_api_key(settings: &Settings, verbose: bool) -> Result<Self> {
```
carlsverre, line 242: **Cleanup — `new` and `new_requiring_api_key` are copy-paste.** They're identical except `Auth::from_env()` vs `Auth::api_key_from_env()`; the `cache_dir()` lookup + `as_deref()` + `build(...)` plumbing is duplicated in both. Collapse to one private `new_with_auth(settings, auth, verbose)` so the cache wiring can't diverge between the two.

tests/support/mod.rs
```diff
@@ -127,21 +127,39 @@ impl MockDocsServer {
...
+    cmd.env("RUST_LOG", "debug");
+
+    // Global settings can't leak (HOME/XDG_CONFIG_HOME aren't forwarded, so the
+    // lookup resolves to nothing). Pin the project settings file to an empty
+    // but existing file so a ./.snouty.toml in the working tree can't leak in
+    // either — empty so it contributes no values, existing so the explicit-path
+    // resolution doesn't error. A test wanting project settings overrides it.
+    let empty_settings =
+        std::path::Path::new(env!("CARGO_TARGET_TMPDIR")).join("empty.snouty.toml");
+    let _ = std::fs::write(&empty_settings, "");
```
carlsverre, line 161: **Test robustness — shared empty `.snouty.toml` written racily to fake "no project settings".** Every `snouty()` does `let _ = fs::write(&empty_settings, "")` to one shared `CARGO_TARGET_TMPDIR/empty.snouty.toml` and pins `SNOUTY_SETTINGS_PATH` to it. Idempotent today (always empty), but it's shared mutable state written from parallel tests with the error swallowed, and it papers over a missing first-class "ignore any project file" mode (which also means the explicit-path-missing branch is never exercised through the normal harness). A sentinel / `--no-settings` opt-out would be cleaner.

jeskew: I don't think we want a `--no-settings` mode that disables reading from settings. Tests that want to use a different project settings file can set their own `SNOUTY_SETTINGS_PATH` or pass a different path in.

carlsverre: agreed

src/main.rs
jeskew: An update from today made settings resolution lazy, so we can safely move the `Settings::new` call earlier in the function and avoid the double match.

PR #130, opened 2026-06-11
fix: runs list robustness and clearer auth errors

### LINKED ISSUE #121: BUG: Panic on broken pipe in `snouty runs list`
### What happened?

`snouty runs list` doesn't gracefully handle broken pipes, e.g. from piping the output into `head`.

### What command did you run?

snouty runs list --limit 1000 | head -n1

### What was the output?

```text
RUN ID                                 STATUS       CREATED AT                     LAUNCHER
The application panicked (crashed).
Message:  failed printing to stdout: Broken pipe (os error 32)
Location: library/std/src/io/stdio.rs:1165

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

### What is your snouty version?

snouty 0.5.0

### LINKED ISSUE #122: BUG: `snouty runs list` incompatible with in-development run schemas, affecting an internal tenant
### What happened?

Testing on an internal tenant, `antithesis.is_ephemeral` had at some point in the past the value `"on"`, likely before the API or schema was stabilized or perhaps as an artifact of a transient bug during development. This leads leads to a hard error shown below, with no incremental results shown.

### What command did you run?

snouty runs list --limit 10000

### What was the output?

```text
snouty runs list --limit 10000
Error:
   0: invalid API response payload: unknown variant `on`, expected `true` or `false` at line 1 column 29570
        ...esis.integrations.type":"none","antithesis.is_ephemeral":"on","run.caller_type":"antithesis_ui"}},{"run_id":"9d5ac27d8f7...
                                                                       ^

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

### What is your snouty version?

snouty 0.5.0

### LINKED ISSUE #129: snouty api usage requires api key. better errors and docs
### What happened?

all of the openapi endpoints other than launch require an API key.
snouty should error with a clear message rather than hitting the API with user/pass or no auth.
docs should be updated to make the requirement clear.

### What command did you run?

snouty runs list

### What was the output?

```text
403 not authorized
```

### What is your snouty version?

latest

carlsverre:
Fixes #122
Fixes #121
Fixes #129

`snouty runs list` hard-errored on runs whose `antithesis.is_ephemeral` was stored as `"on"` rather than `"true"` (#122). The enum is generated from openapi.json by progenitor at build time, so build.rs now patches the generated client to add `serde(alias = "on")`/`alias = "off"` to the true/false variants, normalizing the values on the way in. The patch asserts the expected generated code is present exactly once, so a progenitor upgrade that changes the output shape fails the build instead of silently dropping the aliases.

Piping runs output into something that exits early (`snouty runs list | head -n1`) panicked on the broken pipe (#121). The runs list/show/properties paths now print through a small `outln!`/`out!` macro pair that writes to stdout with fallible `writeln!`/`write!` rather than `println!`, so a closed pipe surfaces as an io::Error instead of a panic, and main treats a BrokenPipe write error as a normal quiet exit. Network errors don't take that path: they surface as reqwest errors whose inner io::Error is not downcastable from the report.

Every API endpoint other than launch rejects basic auth, but snouty would happily send username/password and surface a bare 403 (#129). The runs commands now require ANTITHESIS_API_KEY up front and fail with a message explaining that username/password authentication only works for the launch commands (`snouty launch`, `snouty debug`, `snouty api webhook`), with a suggestion to ask your Antithesis contact for a key. The README auth section, the runs spec, and the AGENTS.md staging notes are updated to match.

Each fix has a regression test: a wiremock test covering on/off parsing, a CLI test that closes snouty's stdout and asserts a clean exit, and new runs spec assertions for the fail-fast auth error.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -175,7 +175,7 @@ impl Auth {
             );
         }
         Err(err.suggestion(
-            "set ANTITHESIS_API_KEY; ask your Antithesis contact for an API key if you \
+            "set ANTITHESIS_API_KEY; ask Antithesis support for an API key if you \
              don't have one",
         ))
     }
--- a/src/api.rs
+++ b/src/api.rs
@@ -154,6 +154,31 @@ impl Auth {
             required_env("ANTITHESIS_PASSWORD")?,
         ))
     }
+
+    /// Like [`Auth::from_env`], but only accepts an API key. The API rejects
+    /// basic auth everywhere except the launch endpoints, so commands that hit
+    /// other endpoints should fail fast with a clear message instead of
+    /// sending a request destined for a 403.
+    fn api_key_from_env() -> Result<Self> {
+        if let Some(api_key) = optional_env("ANTITHESIS_API_KEY")? {
+            return Ok(Self::bearer(api_key));
+        }
+        let has_basic = optional_env("ANTITHESIS_USERNAME")?.is_some()
+            || optional_env("ANTITHESIS_PASSWORD")?.is_some();
+        let mut err = eyre!("missing environment variable: ANTITHESIS_API_KEY");
+        if has_basic {
+            err = err.note(
+                "ANTITHESIS_USERNAME/ANTITHESIS_PASSWORD are set, but this command only \
+                 accepts API key authentication; username/password authentication is only \
+                 supported when launching runs (`snouty launch`, `snouty debug`, \
+                 `snouty api webhook`)",
+            );
+        }
+        Err(err.suggestion(
+            "set ANTITHESIS_API_KEY; ask your Antithesis contact for an API key if you \
+             don't have one",
+        ))
+    }
 }
 
 #[derive(Clone, Debug)]
@@ -180,6 +205,16 @@ impl Config {
             verbose: false,
         })
     }
+
+    /// Like [`Config::from_env`], but only accepts API key authentication.
+    pub fn from_env_requiring_api_key() -> Result<Self> {
+        debug!("loading config from environment (API key required)");
+        Ok(Self {
+            auth: Auth::api_key_from_env()?,
+            tenant: required_env("ANTITHESIS_TENANT")?,
+            verbose: false,
+        })
+    }
 }
 
 pub struct AntithesisApi {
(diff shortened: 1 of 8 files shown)


src/api.rs
```diff
@@ -154,6 +154,31 @@ impl Auth {
...
+            err = err.note(
+                "ANTITHESIS_USERNAME/ANTITHESIS_PASSWORD are set, but this command only \
+                 accepts API key authentication; username/password authentication is only \
+                 supported when launching runs (`snouty launch`, `snouty debug`, \
+                 `snouty api webhook`)",
+            );
+        }
+        Err(err.suggestion(
+            "set ANTITHESIS_API_KEY; ask your Antithesis contact for an API key if you \
+             don't have one",
```
carlsverre: ask antithesis support

src/main.rs
```diff
@@ -109,6 +109,22 @@ async fn main() -> Result<()> {
...
+    // underlying io::Error is not downcastable from the report.
+    match result {
+        Err(err)
+            if err
+                .downcast_ref::<io::Error>()
+                .is_some_and(|e| e.kind() == ErrorKind::BrokenPipe) =>
+        {
+            Ok(())
+        }
+        other => other,
```
carlsverre, line 147: can we somehow sanity check that this doesn't trigger on network errors? relying on downcast semantics feels janky. or perhaps there is a way to check that this error is coming from a write to stdout/stderr?

tests/cli_general.rs
```diff
@@ -866,3 +866,43 @@ fn validate_help_shows_keep_running() {
...
+        .env("ANTITHESIS_TENANT", "testtenant")
+        .env("ANTITHESIS_BASE_URL", &mock.url)
+        .args(["runs", "list"])
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .spawn()
+        .unwrap();
+
+    // Close the read end of stdout so snouty's writes hit a broken pipe.
+    drop(child.stdout.take());
```
carlsverre: is this racy? i'm assuming the snouty process spawns on line 894, so in theory it can complete before we close the pipe. can we create the pipe first, and pre-close stdout?

src/api.rs
carlsverre: Done in ef3cb7f7 — suggestion now reads "ask Antithesis support for an API key" (README updated to match).

src/main.rs
carlsverre: Done in ef3cb7f7 — stdout writes are now tagged at the write site with an explicit `StdoutWriteError` marker type (`runs.rs`), and main only takes the quiet-exit path for that type. Io errors from reqwest or anywhere else can no longer match, regardless of downcast semantics.

tests/cli_general.rs
carlsverre: Done in ef3cb7f7 — the test now creates the pipe itself via `std::io::pipe()` and drops the read end before spawning, so the first write deterministically hits EPIPE.

PR #138, opened 2026-06-13
Improve `snouty runs` output and `--help` quality

carlsverre:
Polishes snouty's human-facing CLI output, help text, and error rendering, driven by the gallery satisfaction reviews. The headline change folds the `runs property` subcommand into `runs properties` and reworks how properties and their examples render; alongside it, every error now prints through one clean report format.

- `runs logs` renders vtime at a fixed 3 decimals, **truncated** (not rounded), so a value copied off-screen into `--begin-vtime` lands on — never past — the line you saw, and the decimal column stays aligned.
- `runs events` searches a run's events for one or more substrings and prints the matching rows. SOURCE auto-sizes to its widest cell, the OUTPUT cell is whitespace-trimmed, and on a real terminal the OUTPUT column is windowed around the first match — centered on it with an ellipsis marking each truncated edge — so the hit you searched for stays visible; piped output is left full.
- `runs properties` renders one table per group (the group is the section heading), with the EXAMPLES count right-aligned and SI-shortened (`2.3k`, `18M`) and counterexamples shown inline as `examples/counterexamples` when present. Ungrouped properties go in a trailing "(ungrouped)" section.
- The `runs property` subcommand is removed and folded into `runs properties`: narrow with `--name` (case-insensitive substring) and/or `--group`, and add `--detail` to expand the matches. Appending a filter is far quicker than swapping the subcommand word, and this drops the fuzzy/ambiguous/not-found resolution entirely (a filter that matches nothing just prints a friendly empty state).
- In `--detail`, event properties show an `Examples` moment table (the HASH/VTIME you paste into `runs logs`); non-event "system" properties — which store data, not moments — show their value under a `Result` label instead (a scalar inline, an object/array inline when it fits the terminal, otherwise pretty-printed JSON). A *failing* non-event property carries its violating value in counterexamples, which `--detail` labels under `Counter-examples` (shown first), distinct from the satisfying `Examples`, so the offending value isn't lost in an unlabelled list. `--detail` errors when combined with `--json`, on both `runs properties` and `runs list`.
- `runs show` prints `Failure Hash` before `Failure VTime` (matching the `runs logs <hash> <vtime>` argument order) and hands over a ready-to-paste `snouty runs logs …` command for incomplete runs, the way the `--web` hint does for the report.
- Errors now render through one report format. User-facing failures (and 4xx API errors) print cleanly — a plain statement of what went wrong plus any `.note()`/`.suggestion()`/`.section()` hints, with no backtrace — while genuine internal faults keep theirs. This replaces the `is_user_error` / `{:#}`-vs-`{:?}` branching in `main` with a single rendering, drops the env-section footer, and emits color only to a TTY (pipes and tests get plain text). The why, the remediation, and the diagnostic detail move out of the message and into notes/suggestions/sections; applied across runs, api, main, params, moment, config, docs, validate, and doctor, with bad args, config layout, missing env vars, missing docs db, and version-gated resources classified as user errors while network/runtime/serialization faults stay internal.
- `--help` is overhauled across every command: output legends and examples on the read/stream commands, a consistent option layout, the global `--json`/`--verbose` flags sorted to the bottom so each command's own flags stay grouped, `--duration` documented as minutes, and the `launch --config` wall trimmed.
- `gen-gallery.py` now renders each command's `--help` next to its default output as "help stories", with an automated check that any column the help names actually appears in the output — so the help can be reviewed for clarity and alignment, not just the command output. A help story whose default command exits non-zero now fails the run (doctor excepted). AGENTS.md gains a `scripts/` section (Python helpers run via `uv`, type-checked with `uvx pyright`), a top-level `CLAUDE.md` points at it, and Python bytecode is gitignored. Its non-event property discovery now recognizes snouty's bare `Result` label (the no-colon form used for object/array values) and probes several candidates instead of bailing on the first, so one oddly-rendering property can no longer abort generation of the whole gallery.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 /target
 .claude/settings.local.json
+
+# Python bytecode cache (e.g. from running scripts/gen-gallery.py)
+__pycache__/
+*.pyc
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,6 +7,7 @@ CLI tool for the Antithesis API. Written in Rust.
 - `specs/` — feature specs
 - `src/` — all source code
 - `tests/` — integration tests
+- `scripts/` — maintenance/dev scripts (Python, run via `uv`)
 - `.github/workflows/` — CI/CD
 
 ## Specs
@@ -56,6 +57,16 @@ lines prefixed with `[!staging]` are skipped (those assert on hardcoded
 mock data); unprefixed lines still run and hit staging. Only read-oriented
 checks run against staging — any spec that would mutate state is gated
 `[!staging]`.
+
+## Scripts
+
+The `scripts/` directory holds Python helpers run via `uv` (each has an inline
+`# /// script` dependency header). After changing `gen-gallery.py`, type-check it
+with pyright and confirm it reports 0 errors before considering the change done:
+
+```
+uvx pyright scripts/gen-gallery.py
+```
 
 ## AI Coding Workflow
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+See [AGENTS.md](AGENTS.md) for project instructions.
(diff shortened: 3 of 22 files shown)


carlsverre:
@claude review

carlsverre:
@claude review

PR #143, opened 2026-06-17
feat: support --run-id on snouty debug

BillGrieser-Antithesis:
## What & why

`snouty debug` now accepts `--run-id` in addition to `--session-id` to identify the moment to recreate. **Exactly one of the two is required.** This lets the antithesis-skills agents launch MVD sessions directly from a run id (which triage already hands off) instead of asking the user to look up a session id. `--session-id` stays supported for moments copied straight from a triage report.

The debugging API models this as a `oneOf` (run-id-identified vs session-id-identified) in the vendored OpenAPI spec, so the generated client exposes an `MvdParams` enum. `launch_mvd_request` builds the matching variant, and `Params::ensure_single_debug_target` enforces the exactly-one rule with a clear error (both → "specify exactly one…"; neither → "specify --run-id or --session-id").

Moment/stdin parsing is unchanged and stays field-agnostic: a `run_id` or `session_id` moment passes through as-is (the triage report can switch from `session_id` to `run_id` without a snouty change).

## Launch/debug response tolerance

The vendored spec documents the `launch` and `launchMvd` success responses as HTTP `202` with a `run_id` body, but the deployed API returns the Antithesis webhook envelope — HTTP `200` with `{"statusCode": 202, "runId": ...}`. Without handling, the generated client rejects the live `200` as `API error: 200 OK`, mis-reporting a successful launch.

`finish_launch` now treats **any 2xx** HTTP status as success and reads the run id from the body (tolerating `run_id`/`runId`). `snouty debug` emits the same `runId`/`statusCode` envelope as `snouty launch` for `--json`.

> `src/openapi.json` is the upstream-owned vendored spec (copied in, not edited here), so this is a client-side tolerance rather than a spec change.

## Also

- Version bump `0.5.0` → `0.6.0`.
- Merged latest `main`.

## Testing

Local `cargo test` (lib unit tests + `tests/cli_*.rs` + `tests/spec.rs`), `cargo clippy --all-targets`, and `cargo fmt --check` all pass. New coverage includes the run-id / session-id / both / neither paths, the HTTP-200 webhook envelope for both launch and debug, and `--run-id` / `--session-id` spec cases in `specs/debug.txt`.

### DIFF
--- a/specs/debug.txt
+++ b/specs/debug.txt
@@ -43,7 +43,7 @@ stderr 'specify --run-id or --session-id.*'
 # Parameters are provided as typed flags. Before sending, the resolved
 #    parameters are printed to stderr with sensitive values redacted. The
 #    target run is identified by --run-id (preferred for new integrations).
-mock-server 202 '{"run_id":"run-123"}'
+mock-server 200 '{"statusCode":202,"runId":"run-123"}'
 snouty debug --input-hash abc123 --run-id 9043254f65c9c65d63fe043a0abfc7fc-53-1 --vtime 1234567890 --description 'debug this moment' --recipients team@example.com
 stderr '"antithesis.debugging.input_hash": "abc123"'
 stderr '"antithesis.debugging.run_id": "9043254f65c9c65d63fe043a0abfc7fc-53-1"'
@@ -52,7 +52,7 @@ stderr '"antithesis.event_description": "debug this moment"'
 stderr '"antithesis.report.recipients": "\[REDACTED\]"'
 
 # --session-id remains supported for backwards compatibility.
-mock-server 202 '{"run_id":"run-123"}'
+mock-server 200 '{"statusCode":202,"runId":"run-123"}'
 snouty debug --input-hash abc123 --session-id d5d1c9cf0e64b5dd69b00e97b07fe3f4-18-1 --vtime 1234567890
 stderr '"antithesis.debugging.session_id": "d5d1c9cf0e64b5dd69b00e97b07fe3f4-18-1"'
 
@@ -63,7 +63,7 @@ stderr '"antithesis.debugging.session_id": "d5d1c9cf0e64b5dd69b00e97b07fe3f4-18-
 #    converted to strings. The moment carries whichever identifier the triage
 #    report provides (session_id today, run_id going forward) and is passed
 #    through unchanged.
-mock-server 202 '{"run_id":"run-123"}'
+mock-server 200 '{"statusCode":202,"runId":"run-123"}'
 stdin moment_run_input.txt
 snouty debug --stdin
 stderr '"antithesis.debugging.run_id": "9043254f65c9c65d63fe043a0abfc7fc-53-1"'
@@ -71,7 +71,7 @@ stderr '"antithesis.debugging.input_hash": "6057726200491963783"'
 stderr '"antithesis.debugging.vtime": "329.8037810830865"'
 
 # A session-id moment (today's triage report format) still works.
-mock-server 202 '{"run_id":"run-123"}'
+mock-server 200 '{"statusCode":202,"runId":"run-123"}'
 stdin moment_input.txt
 snouty debug --stdin
 stderr '"antithesis.debugging.session_id": "f89d5c11f5e3bf5e4bb3641809800cee-44-22"'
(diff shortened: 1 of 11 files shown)


specs/debug.txt
```diff
@@ -8,85 +8,114 @@
...
 stdout 'Debugging session started: run_id run-123'
 
 # With --json, the full response body is printed as pretty JSON.
-mock-server 200 '{"runId":"run-123","statusCode":202}'
-stdin moment_input.txt
+mock-server 200 '{"statusCode":202,"runId":"run-123"}'
+stdin moment_run_input.txt
 snouty --json debug --stdin
-stdout '\s*"runId": "run-123".*'
+stdout '\s*"run_id": "run-123".*'
```
carlsverre: **Stale integration test still asserts camelCase `runId` — `cargo test` fails on this PR.**

This spec correctly flips `--json debug` to snake_case `run_id` to match the new `LaunchMvdResponse { run_id }`. But the equivalent integration test `tests/cli_general.rs:363` (`debug_with_json_flag`) was *not* updated and still asserts:

```rust
.stdout(predicate::str::contains(r#""runId": "run-abc-1-1""#));
```

After the return-type change, `snouty --json debug` now emits `{"run_id": "..."}`, so that assertion no longer matches and `cargo test` reports:

```
test result: FAILED. 33 passed; 1 failed
failures:
    debug_with_json_flag
```

Update that assertion to `"run_id"` (and note that `--json debug` now drops `statusCode` from the output, since `LaunchMvdResponse` only carries `run_id` — worth a deliberate decision if any consumer relied on it).

src/api.rs
```diff
@@ -752,34 +760,82 @@ fn launch_request(params: &Params) -> Result<generated::types::LaunchRequest> {
...
+async fn finish_launch<T>(
+    result: std::result::Result<
+        progenitor_client::ResponseValue<T>,
+        ClientError<generated::types::ErrorResponse>,
+    >,
+    from_body: impl FnOnce(serde_json::Value) -> Result<T>,
+) -> Result<T> {
+    match result {
+        Ok(response) => Ok(response.into_inner()),
+        Err(ClientError::UnexpectedResponse(resp)) if resp.status().is_success() => {
```
carlsverre, line 785: **`finish_launch` only tolerates `UnexpectedResponse` — a documented-202 response with a camelCase body would break `snouty debug`.**

This guard rescues a *non-documented* 2xx: the live `/launch/debugging` webhook currently returns HTTP 200 (which lands in the generated client's `_ => UnexpectedResponse` arm), and `from_body` then repairs it by tolerating both `runId` and `run_id`.

But the generated `launch_mvd` handler matches the documented code strictly:

```rust
match response.status().as_u16() {
    202u16 => ResponseValue::from_response(response).await, // strict deser into LaunchMvdResponse { run_id }
    ...
    _ => Err(Error::UnexpectedResponse(response)),
}
```

`LaunchMvdResponse` requires snake_case `run_id`. If the webhook ever returns the now-documented **202** while keeping the camelCase `{"statusCode":...,"runId":...}` envelope it uses for the 200 case, `from_response` fails with `Error::InvalidResponsePayload` — which is *not* `UnexpectedResponse`, so it skips this arm, falls through to `format_launch_client_error`, and `snouty debug` reports an error even though the session launched.

The PR's own tests encode that camelCase-on-200 / snake_case-on-202 split, so this is latent today, but it's a fragile coupling to the server's exact status code. Consider also routing a 2xx `InvalidResponsePayload` through `from_body`, or making `LaunchMvdResponse` accept `runId`. (`launch_test` is unaffected — its `LaunchResponse` is camelCase, so its documented-202 body parses fine.)

specs/debug.txt
carlsverre: Confirmed fixed in 9b42d33 — `launch_debugging` now returns the `{runId, statusCode}` envelope (consistent with `snouty launch`), the spec assertion is back to `runId`, and `debug_with_json_flag` passes again. Nice call keeping `--json` consistent across both commands rather than splitting the output shape. Resolving. ✅

src/api.rs
carlsverre: Thanks — the restructure handles the real path (HTTP 200, camelCase envelope) cleanly and lines the output up with `snouty launch`. 👍

One precise residual, for the record (non-blocking): the `Ok(response)` arm is only reached when the generated client has *already* parsed the 202 body into `LaunchMvdResponse { run_id }` (snake_case). If the webhook is ever "corrected" to return the documented **202** while keeping the camelCase `{"statusCode":…,"runId":…}` envelope it actually returns today, `send()` fails with `Error::InvalidResponsePayload` — which isn't `UnexpectedResponse`, so it skips both success arms and falls to `format_launch_client_error`, erroring on a launch that actually succeeded.

Same latent server coupling I flagged before; since the live API returns 200 it's not reachable now, so I'll leave this open for your call rather than block on it.

PR #157, opened 2026-06-24
Add `snouty login` command to populate settings file and persist credentials

### LINKED ISSUE #150: Support GitHub Actions OIDC authentication
Antithesis supports [GitHub Actions’ OpenID Connect feature](https://docs.github.com/en/actions/concepts/security/openid-connect) for authenticating to the API (including the launch endpoint), by getting a token with an audience of `antithesis` and attaching it to an `Authorization` header as `GHA <token>`. It’d be nice if Snouty supported this automatically if we’re in a GHA environment, and/or no other authentication method has been explicitly provided.

(Talk to me about enabling this on a tenant for testing, if needed.)

### LINKED ISSUE #156: Add command to setup/update settings
Settings get a bit complicated to hand-manage once OAuth support is integrated, as we want to store tokens in an OS keychain if at all possible. This command would be similar to `aws login` -- users would run this to configure credentials after installation and then infrequently afterwards.

jeskew:
Resolves #156 & resolves #150

This PR adds a new `snouty login` command whose primary purpose is to populate the settings file added in #128.

The PR also adds credential persistence per offline discussions so that `snouty login` can capture those as well. Credentials will be stored in one of two places, listed in order of preference:

1. The system keychain (only supported on MacOS).
2. `$XDG_CONFIG_DIR/snouty/credentials.toml`/`$HOME/.config/snouty/credentials.toml` (only used if one of those two environment variables is set)

There is a `SNOUTY_DISABLE_KEYCHAIN_CREDENTIAL_STORAGE` environment variable that skips keychain access. This environment variable is only meant to be set in the test runner to ensure isolation.

The credentials file (if used) has restricted permissions (`0600`) and is stored within a directory that is specific to the user and *should* have restricted permissions. (If Snouty creates the directory, it will have `0700` permissions, but if the `~/.config/snouty` directory already exists, Snouty will not alter its permissions.)

We don't publish a Windows release artifact, but if we ever do, we should add support for the Windows system keychain and ensure that the file-based option is only available on *nix systems. 

The PR also adds a few dependencies, specifically:

<dl>
<dt><code>rpassword</code></dt><dd>Allows collecting secrets from stdin without echoing them back to the TTY</dd>
<dt><code>keyring-core</code></dt><dd>Cross-platform keychain access platform</dd>
<dt><code>apple-native-keyring-store</code></dt><dd>Keyring bindings for MacOS</dd>
</dl>

Since this PR is updating how we load ambient credentials, it adds the requested support for GH Action OIDC authentication as well.


src/credentials.rs
```diff
@@ -188,12 +226,69 @@ impl Credentials {
...
+fn fetch_github_actions_oidc_credentials(
+    actions_id_url: &str,
+    actions_id_request_token: &str,
+) -> Result<Credentials> {
+    let client = reqwest::blocking::Client::builder().build()?;
+    let token = client
+        .get(format!("{actions_id_url}&audience=antithesis"))
+        .bearer_auth(actions_id_request_token)
+        .send()?
+        .text()?;
```
jeskew: One thing to flag here: this is a **blocking** HTTP request, which I am not sure is a great idea. Since we only make this call if both Github Actions-specific environment variables are set, this _should_ always be a local request that resolves very quickly.

There are a couple of alternatives here if that feels a bit dicy:
1. Make credential resolution async and use the vanilla `reqwest` client. (This can still go awry if the target is unreachable or encounters a connection timeout, but we're implicitly surfacing to the caller (within Snouty, not of Snouty) that resolving credentials may take a while.)
2. Require that users call [`getOidcToken('antithesis')`](https://docs.github.com/en/actions/reference/security/oidc#methods-for-requesting-the-oidc-token) within the workflow definition, then expose the result as an environment variable. This makes everything simpler Snouty-side, but it's more complex to use (users need to know the exact audience name and the separate environment variable name, plus they will need to modify their workflows in at least two places -- once to enable OIDC token access and then again wherever they need to expose the env variable.)

jeskew: Went with option 1 here. This ended up being a pretty small change overall.

src/auth.rs
```diff
@@ -0,0 +1,643 @@
...
+
+        Ok(None)
+    }
+
+    pub(crate) async fn for_ambient_credentials_with_attribution(
+        profile: Option<&str>,
+        allow_basic: bool,
+        offline: bool,
+    ) -> Result<AttributedValue<Self>> {
+        if let Some(from_env) = Self::try_from_env()? {
```
carlsverre, line 149: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Exported `ANTITHESIS_USERNAME`/`ANTITHESIS_PASSWORD` shadow a login-stored API key.**

Env is resolved first here, and `to_result` rejects password auth *terminally* for API-key-only commands, so the keychain / `credentials.toml` (where the stored key lives) are never consulted.

Scenario: a user keeps `ANTITHESIS_USERNAME`/`PASSWORD` exported for `snouty launch`, then runs `snouty login` to store an API key. `snouty runs` calls `new_requiring_api_key` → `for_ambient_credentials(allow_basic=false)`; `try_from_env` returns `Password`, `to_result` errors *"does not accept username/password authentication"* and returns **before** reaching the keychain (L196) or file (L200). The just-stored key is unusable until the env vars are unset.

Consider letting a disallowed credential *kind* fall through to the next source rather than terminating resolution.

src/login.rs
```diff
@@ -0,0 +1,200 @@
...
+    credentials::{Credentials, persist},
+    settings::{Settings, update_settings_in_global_file, validate_tenant_host},
+};
+
+pub async fn cmd_login(
+    tenant: Option<String>,
+    repository: Option<String>,
+    current_settings: Result<Settings>,
+) -> Result<()> {
+    let profile = current_settings.as_ref().ok().and_then(|s| s.profile());
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`snouty --profile X login` silently writes to the *default* profile when settings fail to resolve.**

`profile` is read from the resolved `Settings` (`current_settings…s.profile()`), not the CLI `--profile` flag. `main.rs` moved `profile` into `Settings::resolve` and hands `cmd_login` the `Result`. When the settings file is malformed, an explicitly-requested file is missing, or an existing tenant fails `validate_tenant_host`, `resolve` returns `Err` → `profile` is `None` → `persist(creds, None)` + `update_settings_in_global_file(..., None)` write to the default profile / `_default_` keychain entry.

This fires in exactly the broken-config case that deferring `settings?` was meant to support (login is the *repair* command). Thread the CLI `profile` directly into `cmd_login` instead of recovering it from the settings `Result`.

src/settings.rs
```diff
@@ -240,6 +242,93 @@ impl Settings {
...
+        update_table(&mut target, tenant, repository, base_url, container_engine);
+        let profile_value = Value::Table(target);
+
+        match contents.entry(PROFILE_KEY) {
+            Entry::Vacant(vacant) => {
+                let mut outer = Table::new();
+                outer.insert(profile.to_owned(), profile_value);
+                vacant.insert(Value::Table(outer));
+            }
+            Entry::Occupied(mut occupied) => {
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Updating a named profile replaces the whole `[profile.X]` table instead of merging — silently dropping keys the update doesn't set.**

This `Occupied` arm does `outer.insert(profile, profile_value)`, where `profile_value` is a fresh table built from only the provided fields (login always passes `base_url = None`, `container_engine = None`). Any existing `[profile.X].base_url` / `container_engine` is wiped. The no-profile branch (L276) *merges* into existing contents, so profile writes lose data that default writes preserve.

Scenario: a user has `[profile.prod] base_url = ...` set, runs `snouty --profile prod login`, and loses it. Merge into the existing subtable like the default branch does.

src/login.rs
```diff
@@ -0,0 +1,200 @@
...
+fn prompt_for_value(value_name: &str, previous_value: Option<&str>) -> Result<String> {
+    println!("What {value_name} would you like to use?");
+    if let Some(prev) = previous_value
+        && !prev.is_empty()
+    {
+        println!("(Hit enter to use the previous value of [{prev}])");
+    }
+
+    let mut input = String::new();
+    while input.is_empty() {
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Infinite CPU loop on stdin EOF when there's no previous value.**

`read_line` returns `Ok(0)` at EOF (not an error, so `?` doesn't help); with no previous value `input` stays empty and `while input.is_empty()` never exits — pegging a core.

Repro: `snouty login < /dev/null`, Ctrl-D at the tenant prompt on a fresh `$HOME`, or piping fewer lines than there are prompts. (Verified by running this exact loop against `/dev/null`: 5M+ iterations, never terminated.) Treat `Ok(0)`/EOF as a clean error exit.

specs/login.txt
```diff
@@ -0,0 +1,106 @@
...
+# `login` never contacts the API, so the whole file runs in both mock and
+# staging modes. Each section pins its own temporary $HOME via `isolate-home`
+# so the persisted files do not leak between sections or into the developer's
+# real configuration. Prompts are written to stdout; stdin feeds the answers,
+# one per line, in the order the prompts appear.
+
+# 1. A fresh login with no flags prompts for the tenant, the repository, and the
+#    kind of credentials, then persists them. With nothing stored yet, no
+#    "previous value" default is offered for the tenant or repository.
+isolate-home apikey
```
carlsverre, line 21: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**This spec never actually exercises credential persistence, reuse, or username/password auth — and its comments are inaccurate.**

The credential-prompt answers are mis-numbered relative to the menu (`1 = Skip`, `2 = API key`, `3 = Username/password`):
- `login_apikey.txt`, `login_reuse.txt`, `login_profile*.txt` all answer **`1` (Skip)**, so the trailing `sk-test-key`/`pk-secret` lines are never consumed and `credentials.toml` is never written or read — yet the comments say *"then persists them"* and *"credentials.toml was written in step 1"*.
- `login_password.txt` answers **`2`**, which is **API key**, not username/password — so it stores an API key (`puser`) while the comment claims it *"collects a username and password"*.

Net effect: the leave-blank-to-reuse path and option 3 (real username/password) are entirely untested; a regression in any of them would still pass. The fixtures likely need `2`/`3` where they currently use `1`.

src/auth.rs
```diff
@@ -0,0 +1,643 @@
...
+            return Ok(Some(AttributedValue::EnvironmentVariable {
+                value: Self::for_password(username, password),
+                environment_variable_names: vec![USERNAME_VAR_NAME, PASSWORD_VAR_NAME],
+            }));
+        }
+
+        Ok(None)
+    }
+
+    fn try_from_keychain(profile: Option<&str>) -> Result<Option<AttributedValue<Self>>> {
```
carlsverre, line 97: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Keychain profile lookup has no fallback to the default entry, but the file backend does — so the same config resolves differently on macOS vs Linux.**

`try_from_keychain(Some("prod"))` reads only `construct_keychain_credential_name` = `profile_prod`; on a miss it returns `Ok(None)` with no retry against `_default_`. `try_from_credentials_file_toml` *does* fall back to `[default]` (and has a test for it).

Scenario: `snouty login` on macOS stores to keychain `_default_`, then `snouty --profile prod runs` misses `profile_prod`, misses the file, and fails with "No Antithesis credentials found" — while the identical flow succeeds on Linux. Add the same default fallback to the keychain path.

src/login.rs
```diff
@@ -0,0 +1,200 @@
...
+    println!("What kind of credentials would you like to use?");
+    println!(
+        "1. Skip setup (Select this option if you plan to use environment variables instead of persisted credentials.)"
+    );
+    println!("2. API key");
+    println!("3. Username/password");
+    println!("(Hit enter to use the default value of [{default_selection}])");
+
+    let mut input = String::new();
+    while AuthSetupType::try_from_str(&input).is_none() {
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`prompt_for_auth` never clears `input` between reads, so an invalid selection accumulates and can never become valid.**

`read_line(&mut input)` *appends*. Enter `x` then `1` → `input` becomes `x1`, which never parses — an interactive typo is unrecoverable. Combined with EOF it's an infinite loop: `printf 9 | snouty login ...` leaves `input = "9"`, then `read_line` returns `Ok(0)` forever while `try_from_str("9")` stays `None`. Clear `input` each iteration (or read into a fresh `String`).

src/credentials.rs
```diff
@@ -0,0 +1,643 @@
...
+        let credential = match Entry::new("snouty", credential_name.as_str()) {
+            Ok(cred) => Ok(cred),
+            // A NoDefaultStore error indicates that the version of initialize_credential_store() selected by the compiler was a no-op
+            Err(keyring_core::Error::NoDefaultStore) => return Ok(None),
+            Err(other) => Err(other),
+        }?;
+
+        if let Ok(persisted) = credential.get_password() {
+            return Ok(Some(AttributedValue::Keychain {
+                value: serde_json::from_str::<PersistableCredentials>(&persisted)?
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**A corrupt/foreign keychain entry aborts the entire resolution chain instead of falling through.**

This `?` propagates a `serde_json` parse failure on a *successfully read* entry, so a value under service `snouty` name `_default_`/`profile_X` that isn't valid `PersistableCredentials` JSON (schema drift, partial write, manual edit, or another tool's entry) bubbles out of `for_ambient_credentials_with_attribution` — a good `credentials.toml` or OIDC token is never tried. Every non-env command then fails with an opaque JSON error until the entry is cleared manually. Consider treating an unparseable entry as "absent" (fall through) rather than fatal.

src/login.rs
```diff
@@ -0,0 +1,200 @@
...
+            "2" => Some(AuthSetupType::ApiKey),
+            "3" => Some(AuthSetupType::Password),
+            _ => None,
+        }
+    }
+}
+
+async fn prompt_for_auth(profile: Option<&str>) -> Result<Option<Credentials>> {
+    let previous_value =
+        Credentials::for_ambient_credentials_with_attribution(profile, true, false).await;
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`snouty login` can make a live 10s GitHub Actions OIDC network call just to pick a menu default — contradicting "login never contacts the API".**

This resolves ambient credentials with `offline = false`. On a runner with `id-token: write` (`ACTIONS_ID_TOKEN_REQUEST_URL`/`TOKEN` set) and no other creds, it hits GitHub's token endpoint (up to `OIDC_REQUEST_TIMEOUT` = 10s, fresh reqwest client), then throws the token away (it maps to default `1`/Skip and OIDC can't be persisted). Pass `offline = true` — the default should come from what's already persisted locally.

src/auth.rs
```diff
@@ -0,0 +1,643 @@
...
+
+#[cfg(not(target_os = "macos"))]
+pub fn initialize_credential_store() -> Result<()> {
+    // pass
+    Ok(())
+}
+
+pub(crate) fn persist(credentials: Credentials, profile: Option<&str>) -> Result<()> {
+    let persistable = credentials.convert_to_peristable_credentials()?;
+    match try_persist_to_keychain(&persistable, profile) {
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`persist` writes to keychain *or* file but never reconciles them, so a stale/rotated secret can be left in `credentials.toml`.**

Keychain-first: on success it returns `Ok(Some(()))` and never touches the file. Scenario: an API key was written to `credentials.toml` (Linux, or an older build), the user rotates it and re-runs `snouty login` on macOS with the keychain enabled → the new key lands in the keychain, the **old (possibly revoked)** key stays in `credentials.toml`. If the keychain later becomes unavailable (or `SNOUTY_DISABLE_KEYCHAIN_CREDENTIAL_STORAGE` is set), resolution silently falls back to the stale secret. Consider clearing the other backend on persist.

src/doctor.rs
```diff
@@ -263,10 +301,9 @@ fn collect_checks(settings: &Settings) -> Vec<Check> {
     checks.push(repository_check(settings.repository()));
 
     // Authentication (environment-only by design).
-    checks.extend(auth_checks(
-        env_set("ANTITHESIS_API_KEY"),
-        env_set("ANTITHESIS_USERNAME"),
-        env_set("ANTITHESIS_PASSWORD"),
+    checks.extend(authn_checks(
+        Credentials::for_ambient_credentials_with_attribution(settings.profile(), true, offline)
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`snouty doctor` resolves credentials twice per run.**

`collect_checks` resolves here (keychain read + `credentials.toml` parse, and — when `!offline` — the GHA OIDC HTTP fetch), then `cmd_doctor` builds `AntithesisApi::new_requiring_api_key` (L410) which resolves *again*: a second keychain read, second file parse, second reqwest-client build + OIDC exchange. Resolve once and reuse the `AttributedValue` for both the check and the client.

src/main.rs
```diff
@@ -94,10 +96,14 @@ async fn run(cli: Cli) -> Result<()> {
         eprintln!("warning: --json has no effect for `snouty {name}`");
     }
 
+    if let Err(err) = initialize_credential_store() {
```
carlsverre, line 100: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**`initialize_credential_store()` runs for *every* command, including ones that never touch credentials.**

On macOS this builds the keychain `Store` before dispatch even for `version`/`completions`/`docs`/`update`, and prints `warning: Could not initialize system keychain credential storage: ...` to stderr on those commands if init fails (e.g. a headless/SSH session with a locked keychain). Initialize lazily on the credential-resolving path (login + the API builders) instead.

carlsverre (commented):
🤖 **Automated code review by Claude Code (`/code-review`, xhigh effort) — posted on Carl's behalf, not authored by him.**

15 findings (14 confirmed, 1 plausible) are attached inline, ranked roughly most-severe first. Themes: credential **resolution precedence** (exported env shadowing a stored key, a disallowed kind terminating resolution, keychain-vs-file fallback divergence), the `--profile` flag being dropped when settings fail to resolve, profile-table data loss on re-login, two `snouty login` input loops that can hang, and a login spec that doesn't actually cover credential persistence. Please treat each as a suggestion to verify, not as gospel.

src/auth.rs
```diff
@@ -0,0 +1,643 @@
...
+    mkdir(&settings_dir, true, 0o700)?;
+    let mut temp = NamedTempFile::new_in(&settings_dir)?;
+    temp.write_all(toml::to_string_pretty(&current_contents)?.as_bytes())?;
+
+    temp.persist(&path)?;
+
+    Ok(())
+}
+
+fn parse_credentials_file_toml(contents: String, path: &Path) -> Result<CredentialsFile> {
```
carlsverre, line 494: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**A malformed `credentials.toml` can leak secret fragments to stderr via the TOML parse error.**

`toml`'s parse-error `Display` includes the offending source line (with a caret); `wrap_err` keeps it in the chain, and `main.rs` renders errors with `{report:?}`. If a broken line sits on/next to `api_key = "sk-..."` / `password = "..."`, part of the secret is echoed. The trigger is a hand-corrupted file (atomic writes prevent partial writes), so severity is bounded — but this is a secret-handling defect on a file that's `0600` precisely to stay private. Consider not surfacing raw parse output for the credentials file. *(Plausible — depends on the malformed line containing the secret.)*

src/auth.rs
```diff
@@ -0,0 +1,643 @@
...
+impl Credentials {
+    pub(crate) fn for_api_key(api_key: String) -> Self {
+        Self::ApiKey(ApiKeyCredentials { api_key })
+    }
+
+    pub(crate) fn for_password(username: String, password: String) -> Self {
+        Self::Password(PasswordCredentials { username, password })
+    }
+
+    fn try_from_env() -> Result<Option<AttributedValue<Self>>> {
```
carlsverre, line 77: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Partial basic env (only `ANTITHESIS_USERNAME` *or* only `PASSWORD`) is now silently ignored, dropping the old targeted diagnostic.**

`try_from_env` requires *both* halves and returns `None` otherwise, so a user who mistypes `ANTITHESIS_PASSWORD` no longer gets the deleted `Auth::from_env`'s *"missing environment variable: ANTITHESIS_PASSWORD"* — instead resolution either falls through to a stored credential (silently authenticating as a *different* identity) or reports the generic "No Antithesis credentials found". The deleted `auth_partial_basic_errors_like_nothing_set` test also removed coverage here.

src/settings.rs
```diff
@@ -240,6 +242,93 @@ impl Settings {
...
 }
 
+pub(crate) fn update_settings_in_global_file(
+    tenant: Option<String>,
+    repository: Option<String>,
+    base_url: Option<String>,
+    container_engine: Option<String>,
+    profile_to_update: Option<&str>,
+) -> Result<()> {
+    let settings_dir = global_settings_dir().ok_or_eyre("Could not determine global settings directory. Ensure either $XDG_CONFIG_DIR or $HOME is set.")?;
```
carlsverre: > 🤖 _Automated review comment generated by Claude Code (`/code-review`) — posted on Carl's behalf, not authored by him._

**Error message names the wrong env var: it says `$XDG_CONFIG_DIR`, but snouty only reads `XDG_CONFIG_HOME`.**

`global_settings_dir` reads `XDG_CONFIG_HOME` (settings.rs:57), so a user who follows this guidance and sets `XDG_CONFIG_DIR` still fails. The same wrong name is duplicated in `credentials.rs:367`. Should read `XDG_CONFIG_HOME`.

src/auth.rs
jeskew: This is intentional and desired. There are two hierarchies at play here:

1. A hierarchy of credentials, where API keys are preferred over username/password auth, and
2. A hierarchy of input sources, where we try to prioritize sources that more immediately capture the user's intent. I.e., a command line argument is more localized and therefore a stronger signal than an environment variable, which is in turn more localized than a keychain entry or credentials file.

The second hierarchy is I think more important to prioritize for user ergonomics.

src/login.rs
jeskew: Addressed in [8226480](https://github.com/antithesishq/snouty/pull/157/commits/8226480f5e57eb0bbee258aced1a51197b13dc04)

src/settings.rs
jeskew: Addressed in [46308ef](https://github.com/antithesishq/snouty/pull/157/commits/46308efc2de584391a85f9223430a296b43a6c81)

src/login.rs
jeskew: Addressed in [fc69431](https://github.com/antithesishq/snouty/pull/157/commits/fc69431f54681405524eebebd9ee137165b7260f)

specs/login.txt
jeskew: Addressed in [63ad36f](https://github.com/antithesishq/snouty/pull/157/commits/63ad36fb2095e94d119a88e158bed321d78bf563)

src/auth.rs
jeskew: Addressed in [3f435d8](https://github.com/antithesishq/snouty/pull/157/commits/3f435d8c9d7563f93787ab13729be259a4da4cab)

For consistency with snouty settings, an entry in the credentials file under the specified profile is given higher precedence than the default keychain entry. This shouldn't occur under real conditions -- users are either using MacOS, in which case `snouty login` writes to the system keychain and never touches `~/.config/snouty/credentials.toml`, or they are using a different operating system, in which case snouty never touches the keychain. A user could run into this if they manually create a credentials file on a MacOS system, but that seems unlikely to happen.

src/login.rs
jeskew: Mooted by the change in [fc69431](https://github.com/antithesishq/snouty/pull/157/commits/fc69431f54681405524eebebd9ee137165b7260f)

src/credentials.rs
jeskew: Addressed in [5406628](https://github.com/antithesishq/snouty/pull/157/commits/54066281b1ed0d1c00a37455028664f7019d55d9)

src/login.rs
jeskew: Addressed in [b9331c7](https://github.com/antithesishq/snouty/pull/157/commits/b9331c72bfc496960c7017cc43a1502d8b534fff)

src/auth.rs
jeskew: Addressed in [cc97031](https://github.com/antithesishq/snouty/pull/157/commits/cc97031960eae38189e1a2a0aab3b168603b553a)

As discussed [above](https://github.com/antithesishq/snouty/pull/157#discussion_r3532104541), I don't think a system that is using a mixture of keychain and credential files is likely to occur.

src/doctor.rs
jeskew: Because [b9331c7](https://github.com/antithesishq/snouty/pull/157/commits/b9331c72bfc496960c7017cc43a1502d8b534fff) makes the first resolution only check for environment variables/files, I'm OK with the redundant effort here. The OIDC token fetch will only happen on the `AntithesisApi::new_requiring_api_key` call (if ever), and `snouty doctor` is **not** a "hot path" command.

src/main.rs
jeskew: I think this is OK. Snouty otherwise needs to make that call in each command that may use credentials, which leads to a fair amount of duplication.

src/auth.rs
jeskew: In the proposed case, the user has permission to read the file, and Snouty will not be adding any secrets to bash history or anything like that. I think the utility of seeing why their TOML file is invalid outweighs the potential harm of showing the contents of a file the user has permission to `cat`.

src/settings.rs
jeskew: Addressed in [f21fd69](https://github.com/antithesishq/snouty/pull/157/commits/f21fd69aec04d6d067f94fa506464fcc0d6f2b28)

src/auth.rs
jeskew: In this case, the user has not supplied basic auth credentials. We could add a warning about it, but I think it would need to be in addition to the checks already in place.

carlsverre:
@claude review

specs/login.txt
```diff
@@ -0,0 +1,248 @@
...
+# 3. Values supplied as flags are not prompted for: with --tenant and
+#    --repository given, only the credential prompt is shown.
+isolate-home flags
+stdin login_flags.txt
+snouty login --tenant mytenant --repository myrepo
+stdout 'What kind of credentials would you like to use\?.*'
+! stdout 'What Antithesis tenant would you like to use'
+! stdout 'What container repository would you like to use'
+
+# 4. Selecting option 2 collects a username and password instead of an API key.
```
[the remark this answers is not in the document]
carlsverre: Fixed in babc6da — `specs/login.txt` now reads "Selecting option 3 collects a username and password", matching the 1=Skip / 2=API key / 3=Username-password menu.

src/auth.rs
```diff
@@ -0,0 +1,724 @@
...
+
+#[derive(Clone, Debug)]
+pub enum AuthenticationInfo {
+    Static(Credentials),
+    GithubActionsOidc {
+        url: String,
+        request_token: String,
+        cached: Arc<OnceCell<Credentials>>,
+    },
+}
```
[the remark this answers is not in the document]
carlsverre: Addressed in 452712c — `AuthenticationInfo` now has a hand-written `Debug` impl that renders `request_token` as `[REDACTED]`, matching the sibling credential types.

src/login.rs
```diff
@@ -0,0 +1,224 @@
...
+            _ => {
+                return Err(eyre!(
+                    "Exiting login command without completing per user request."
+                ));
+            }
+        }
+    }
+
+    let profile_to_use =
+        profile.or_else(|| current_settings.as_ref().ok().and_then(|s| s.profile()));
```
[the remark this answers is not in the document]
carlsverre: Addressed in 452712c — `cmd_login` now falls back to `env::var(ANTITHESIS_PROFILE_ENV_VAR_NAME)` directly when `--profile` is absent, so the env-var-derived profile survives a failed settings resolve.

jeskew:
@claude review

src/login.rs
```diff
@@ -0,0 +1,226 @@
...
+        io::stdin().read_line(&mut input)?;
+        match input.trim() {
+            "1" => {}
+            _ => {
+                return Err(eyre!(
+                    "Exiting login command without completing per user request."
+                ));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jeskew: Addressed in ecc3ea9

src/login.rs
```diff
@@ -0,0 +1,226 @@
...
+    };
+    validate_tenant_host(&tenant_to_use)?;
+
+    let repository_to_use = match repository {
+        Some(arg_value) if !arg_value.is_empty() => arg_value,
+        Some(_) | None => prompt_for_value(
+            "container repository",
+            current_settings.as_ref().ok().and_then(|s| s.repository()),
+        )?,
+    };
```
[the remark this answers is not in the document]
jeskew: Addressed in ecc3ea9

jeskew:
@claude review

src/cli.rs
```diff
@@ -246,6 +246,38 @@ Examples:
...
+        repository: Option<String>,
+    },
+}
+
+fn validate_non_empty(value: &str) -> Result<String, String> {
+    if value.trim().is_empty() {
+        Err("Value may not be empty or whitespace".to_owned())
+    } else {
+        Ok(value.to_owned())
+    }
```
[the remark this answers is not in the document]
jeskew: Addressed in [fd5576e](https://github.com/antithesishq/snouty/pull/157/commits/fd5576e87981571562f659335dc5a2905303e0ff)

src/login.rs
```diff
@@ -0,0 +1,281 @@
...
+
+fn prompt_for_username_password(
+    previous_username: Option<&str>,
+    previous_password: Option<&str>,
+) -> Result<PersistableCredentials> {
+    Ok(PersistableCredentials::Password {
+        username: prompt_for_value("username", previous_username)?,
+        password: prompt_for_sensitive_value("password", previous_password)?,
+    })
+}
```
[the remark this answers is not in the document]
jeskew: Addressed in [b3d1e18](https://github.com/antithesishq/snouty/pull/157/commits/b3d1e18648def95ef053c16c395a75d4bcdc9c81)

PR #158, opened 2026-06-25
Allow arbitrary additional headers to be set

jeskew:
For consumption by proxies. Leaving this undocumented because it's not a stable part of the Snouty interface.


carlsverre:
this lgtm but plz add a few spec tests to prevent regression.

PR #162, opened 2026-07-02
validate: fail when compose references env vars absent from the Antithesis environment

### LINKED ISSUE #160: Snouty validate should clear the user's environment variables when running the test
an Antithesis user ran into an issue where snouty validate succeeded, and so they assumed things would work in Antithesis. The issue was their docker compose yaml referenced some env vars, and so when their config image did not contain the .env, the Antithesis run failed : (

carlsverre:
`snouty validate` ran docker-compose with the full shell environment, so a docker-compose.yaml that interpolates `${VAR}` could pass locally by drawing the value from the user's shell — while the Antithesis run, which is hermetic and has none of those variables, later failed with an empty (or missing) value. Validation gave false confidence.

validate now checks, before starting anything, that every interpolation variable the compose file references will resolve in the Antithesis environment: from a `.env` file baked into the config image or an inline default. It resolves the compose file under a scrubbed environment and reads which variables docker-compose itself reports as unresolved — its `variable is not set` warnings and `required variable ... is missing` errors — so all dotenv and default semantics (BOM stripping, empty defaults, quoting) are handled by compose rather than re-derived. Any unresolved variable (excluding `ANTITHESIS_*`, which the platform injects) fails validation and is named, with required (`${VAR:?}`) ones annotated.

`--allow-unresolved-env` downgrades the failure to a warning for anyone who bakes the values in some other way. The check runs before the config is otherwise resolved, and is compose-only; Kubernetes configs are unaffected (documented, and guarded by a spec).

An `env_file:` whose path points outside the config dir is out of scope for now; that surfaces clearly on the Antithesis side.

fixes #160

### DIFF
--- a/specs/validate.txt
+++ b/specs/validate.txt
@@ -31,6 +31,10 @@ stdout '.*keep-running.*'
 snouty validate --help
 stdout '.*manifests/.*'
 
+# Help text mentions --allow-unresolved-env.
+snouty validate --help
+stdout '.*allow-unresolved-env.*'
+
 -- empty/.keep --
 -- yml_dir/docker-compose.yml --
 -- ambiguous/docker-compose.yaml --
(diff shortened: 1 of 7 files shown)


carlsverre:
came up with a better solution that doesn't involve a very fragile parser.

PR #164, opened 2026-07-08
Add support for a Snouty-specific proxy setting

jeskew:
Add a configuration setting (`https_proxy` in a settings file or the `ANTITHESIS_HTTPS_PROXY` environment variable) that directs Snouty to forward its requests to Antithesis through a proxy.

This was previously supported by some commands via the `HTTP(S)_PROXY`, as reqwest will use proxy addresses specified there unless an explicit proxy is supplied in code. Those environment variables are  also honored by podman/docker, which some commands invoke on the user's behalf. So if a user wants *everything* to go through a proxy, they can continue to use `HTTP(S)_PROXY` to configure reqwest, podman, and docker; if they want to just proxy Antithesis traffic, `ANTITHESIS_HTTPS_PROXY` will be honored by Snouty but not by any other tools. If the user has set both, `ANTITHESIS_HTTPS_PROXY` takes precedence.

### DIFF
--- a/specs/proxy.txt
+++ b/specs/proxy.txt
@@ -1,0 +1,41 @@
+# HTTPS proxy support
+#
+# When ANTITHESIS_HTTPS_PROXY (or the `https_proxy` setting) is set, snouty
+# routes API requests through that proxy. It builds the client with
+# `Proxy::all`, so both http and https targets are proxied.
+#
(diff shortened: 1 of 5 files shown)


carlsverre (approved):
this lgtm. the 1000 variants of mock servers we're accumulating in spec.rs probably could use a clean up at some point, but not in this diff. thanks for taking point on this!

PR #172, opened 2026-07-13
Support the `docker compose` plugin, and split Compose out of container.rs

carlsverre:
snouty always ran the standalone `docker-compose` binary. On machines with a
modern Docker that only ships the `docker compose` CLI plugin (no standalone
binary), that failed with "docker-compose not found". This adds plugin support,
and reworks how snouty models and drives Compose along the way.

## Compose CLI resolution

snouty now uses whichever Compose v2 form is present:

- the standalone `docker-compose` binary, preferred when present (unchanged for
  existing setups); or
- the `docker compose` plugin, used only when `docker` is real Docker (not
  podman in disguise) and reports v2.

Version is read with `docker compose version --short` and gated on major >= 2,
so Compose v1 is rejected with a clear message. When neither form is usable the
error names both and keeps the specific reason (podman in disguise, a v1 plugin,
docker missing).

The two forms are an enum, so a binary can't be paired with the wrong
invocation. The resolved binary is an absolute path so it still runs under the
scrubbed environment used for the divergence check. That check clears the
environment, which also hid the plugin from the docker CLI when the plugin lived
in a non-default `DOCKER_CONFIG` directory, producing a false "won't resolve in
Antithesis"; snouty now restores `DOCKER_CONFIG` for the plugin form.

## Refactor

`container.rs` had grown to ~3400 lines mixing two concerns. This splits them up:

- All Docker Compose code moved into a new `compose` module. `container.rs` is
  down to ~1400 lines.
- Generic child-process helpers (`output_with_timeout`, `ProcessGroupChild`)
  moved into a `process` module, and `copy_dir_recursive` into a `util` module.
  Neither is Compose- or engine-specific.
- `DockerCompose` is now the one Compose API. It's bound to a single config
  directory (methods take only the overlay) and no longer holds a
  `ContainerRuntime` — the image operations that need one take it as an argument.
- Image architecture is an `Architecture` enum rather than a bare string.

## Other changes

- `snouty doctor` reports the real engine (podman/docker) instead of the
  invoking binary name, matching what `launch`/`validate` announce. Before this,
  a podman-in-disguise setup showed as "docker" in doctor but "podman" in
  launch/validate.
- The `--keep-running` teardown hint prints the actual command
  (`docker compose` vs `docker-compose`) with absolute paths, so it works from
  any directory.
- README: the requirement now covers either Compose form.
- AGENTS.md: unit tests for small, simple internal functions are a judgement
  call rather than a blanket requirement.

## Testing

`cargo build`, `cargo clippy`, and the test suite pass.

Ran snouty by hand across docker (plugin in the system, user, and custom
`DOCKER_CONFIG` dirs), standalone `docker-compose`, native podman, and
podman-in-disguise, checking resolution, the engine announce, and the divergence
check in each. The plugin-discovery fix was confirmed against a non-default
`DOCKER_CONFIG`, where the bare environment scrub fails to find the plugin and
the restore fixes it.


src/container.rs
```diff
@@ -487,20 +481,146 @@ fn stderr_indicates_missing_source(stderr: &str) -> bool {
...
+/// (`program = docker-compose`, no prefix) and the Docker CLI plugin
+/// (`program = docker`, prefix `compose`). snouty drives whichever it finds;
+/// both are genuine Compose v2, whose features snouty depends on. The program
+/// is always an absolute path so the invocation survives the `env_clear()` in
+/// the hermetic render, where `PATH` is gone.
+#[derive(Clone, Debug)]
+pub struct ComposeCommand {
+    program: PathBuf,
+    prefix: &'static [&'static str],
+}
```
carlsverre: We only support two forms: `docker-compose` and `docker compose`. IMO this is much clearer as an enum that clearly encodes the two cases.

src/container.rs
```diff
@@ -487,20 +481,146 @@ fn stderr_indicates_missing_source(stderr: &str) -> bool {
...
+            }
+            Err(e) => Err(eyre!("failed to run `{} version`: {e}", self.display())),
+        }
+    }
+
+    /// A fresh [`Command`] for this invocation with the compose prefix applied.
+    fn command(&self) -> Command {
+        let mut cmd = Command::new(&self.program);
+        cmd.args(self.prefix);
+        cmd
```
carlsverre: this seems brittle - a downstream operation may call args which presumably will override the prefix set here. IMO it may be cleaner to write a function for each docker compose command you want to execute (e.g. `up`, `version`, `config`) exposing inputs and arguments as clear arguments and return values. turn docker compose into an api

src/container.rs
```diff
@@ -572,7 +650,7 @@ fn is_compose_v2_version(output: &str) -> bool {
 pub struct DockerCompose<'a> {
```
carlsverre: IMO merge this with ComposeCommand. it's one concept.

also why does this type need the container runtime?

carlsverre (commented):
consider moving all the docker compose stuff into it's own module. container.rs is getting massive

src/container.rs
carlsverre: Done — `ComposeCommand` (program + free-form prefix) is now the `ComposeCli` enum with `Standalone(PathBuf)` / `Plugin(PathBuf)`. The prefix is derived from the variant, so a program paired with the wrong prefix is unrepresentable. (commit `49ee224`)

src/container.rs
carlsverre: Addressed. The program + compose prefix are now fixed by the enum variant and can't be clobbered by later args. Beyond that, all compose subcommands go through a single `command(overlay, sub)` builder (no method hand-assembles an invocation), the two identical `up`/`logs` spawns collapsed into `spawn_inherited`, and the public surface is a typed method per compose command (`contents`/`ps`/`config_json`/`up_detached`/`down`/`pin_images`/…). (commits `49ee224`, `21aee00`)

src/container.rs
carlsverre: Merged — there is now one concept: `DockerCompose`. `ComposeCli` is a private impl detail; callers use `DockerCompose::resolve(rt, config)` (wired driver) or `DockerCompose::probe()` (diagnostic name+version).

On the runtime: you were right, it didn't belong. `DockerCompose` no longer stores a `ContainerRuntime` (and lost its lifetime param). Image ops take `&dyn ContainerRuntime` explicitly (`pin_images(rt, registry)`), `discover_scripts` uses the `rt` already in scope, and `find_remote_pin` became a free fn since it only needs `rt`. (commits `49ee224`, `21aee00`)

carlsverre:
Pushed three commits addressing the review. Ready for another look.

**`49ee224` — Compose forms as an enum; `DockerCompose` is the one API**
- `ComposeCli` enum (`Standalone`/`Plugin`) — invalid program+prefix combos are unrepresentable.
- Merged `ComposeCommand` into `DockerCompose`; dropped the stored `ContainerRuntime` (and the lifetime param). Image ops take `rt` explicitly.

**`5e341c0` — Extract the Compose integration into its own module** (your "container.rs is getting massive" comment)
- New `src/compose.rs` (~2000 lines): `ComposeCli`/`DockerCompose`, config/ps parsing, image availability + architecture checks, config-image staging, and their tests.
- `container.rs`: ~3400 → ~1400 lines, keeps the runtime/registry primitives and exposes a few shared helpers as `pub(crate)`.
- Pure move — tests pass unchanged.

**`21aee00` — Simplify the Compose API surface**
Before pushing I ran an adversarial review focused specifically on how we drive Compose codebase-wide. It surfaced four worth-doing simplifications, all applied:
- **Bound `ComposeConfig` to the handle** (`resolve(rt, config)`). Config was a constant threaded through ~11 methods; it's just a `PathBuf`, so owning it adds no lifetime. Call sites are now `compose.ps(overlay)`, `compose.contents(None)`, `compose.pin_images(rt, registry)`. `overlay` stays a param — it genuinely varies within a run.
- **Fixed a leaky abstraction**: validate.rs hand-reassembled the `… down` hint from `docker_host()` + `display()` + a duplicated `-f docker-compose.yaml`, which could drift from what `down()` runs. Now `compose.down_hint()`, built from the same internals; `display()`/`docker_host()` are private. Locked with a unit test.
- **One `command(overlay, sub)` builder** for all subcommands + deduped the two identical `up`/`logs` spawns into `spawn_inherited`.
- `find_remote_pin` → free function (never used `self`).

426 tests pass, clippy clean, `snouty doctor` verified on a plugin-only machine.

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+fn compose_file_args(overlay: Option<&Path>) -> Vec<String> {
+    let mut args = vec!["-f".to_string(), "docker-compose.yaml".to_string()];
+    if let Some(path) = overlay {
+        args.push("-f".to_string());
+        args.push(path.display().to_string());
+    }
+    args
+}
+/// The standalone Docker Compose v2 binary name.
+const DOCKER_COMPOSE: &str = "docker-compose";
```
carlsverre: inline

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+/// Build the `-f` flags for `compose` subcommands: always
+/// `-f docker-compose.yaml`, plus `-f <overlay>` if an overlay was provided.
+fn compose_file_args(overlay: Option<&Path>) -> Vec<String> {
+    let mut args = vec!["-f".to_string(), "docker-compose.yaml".to_string()];
+    if let Some(path) = overlay {
+        args.push("-f".to_string());
+        args.push(path.display().to_string());
+    }
+    args
+}
```
carlsverre: inline, no tests needed

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+        Ok(candidate)
+    }
+
+    /// The invocation as a user would type it, for user-facing hints.
+    fn display(&self) -> &'static str {
+        match self {
+            ComposeCli::Standalone(_) => "docker-compose",
+            ComposeCli::Plugin(_) => "docker compose",
+        }
+    }
```
carlsverre: impl Display?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+    /// undiscoverable. Restore `DOCKER_CONFIG` only — docker machinery, not a
+    /// value users interpolate into compose files — so lookup works without
+    /// reintroducing `$HOME` as a `${VAR}` interpolation source.
+    fn restore_discovery_env(&self, cmd: &mut Command) {
+        if let ComposeCli::Plugin(_) = self
+            && let Some(config_dir) = docker_config_dir()
+        {
+            cmd.env("DOCKER_CONFIG", config_dir);
+        }
+    }
```
carlsverre: this is code smell. IIRC we only use this for docker compose config, so I'd just inline this behavior there. this function is otherwise way too weird (modifying a Command based on this type... but no link between this type and the command)

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+            Err(e) => Err(eyre!("failed to run `{name} version`: {e}")),
+        }
+    }
+}
+/// Whether a `compose version` banner identifies Docker Compose v2. Both the
+/// standalone binary and the CLI plugin print "Docker Compose version ..." for
+/// v2; Compose v1's "docker-compose version ..." (hyphenated) does not match.
+fn is_compose_v2_version(output: &str) -> bool {
+    output.to_lowercase().contains("docker compose")
+}
```
carlsverre: is this the best way to detect compose v2? we can't use a version number?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+        if let ComposeCli::Plugin(_) = self
+            && let Some(config_dir) = docker_config_dir()
+        {
+            cmd.env("DOCKER_CONFIG", config_dir);
+        }
+    }
+
+    /// Return the version banner, erroring when the command fails to run or
+    /// reports a non-v2 (v1) banner.
+    fn version(&self) -> Result<String> {
```
carlsverre: version shouldn't *also* handle erroring on v1. that's conflating responsibility. return a version type enum from this fn (i.e do detection) and let the caller decide if it is ok with v1 version of the command.

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+/// The docker CLI config directory (which holds `cli-plugins/`): `$DOCKER_CONFIG`
+/// if set, else `$HOME/.docker`. Read from snouty's own (un-scrubbed) environment.
+/// `None` when neither is set, in which case plugin lookup falls back to the
+/// system directories.
+fn docker_config_dir() -> Option<PathBuf> {
+    if let Some(dir) = std::env::var_os("DOCKER_CONFIG") {
+        return Some(PathBuf::from(dir));
+    }
+    std::env::var_os("HOME").map(|home| PathBuf::from(home).join(".docker"))
+}
```
carlsverre, line 194: maybe we should try to resolve the docker compose plugin path first and then setup the env to resolve the plugin in the scrubbed env?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+/// Docker daemon, or an explicit `DOCKER_HOST` inherited from the environment).
+///
+/// Image operations (tag, push, pin) are not compose's concern — the methods
+/// that need a container engine (e.g. [`pin_images`](Self::pin_images)) take a
+/// [`ContainerRuntime`] argument rather than this type owning one.
+pub struct DockerCompose {
+    cli: ComposeCli,
+    docker_host: Option<String>,
+    config: ComposeConfig,
+}
```
carlsverre, line 208: why do we have DockerCompose and ComposeCli? Why are they not one type?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+    /// none is available. Resolution runs `version` internally to confirm v2 —
+    /// callers that want the banner call [`version`](Self::version) on the
+    /// result.
+    ///
+    /// Prefers the standalone `docker-compose` binary when it is present and
+    /// genuinely v2 (the historical contract). Otherwise falls back to the
+    /// `docker compose` CLI plugin — but only when `docker` is real Docker,
+    /// never podman in disguise, whose compose provider may not implement the
+    /// v2 features snouty relies on.
+    fn resolve() -> Result<ComposeCli> {
```
carlsverre, line 119: the code may be made simpler if we always use `docker compose` (plugin) if the current container engine is docker (and not podman in disguise) (and if the version check passes).

reason about both ways and pick the simpler one.

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+
+    /// Spawn a long-running compose subcommand (`up`, `logs`) with inherited
+    /// stdio, in its own process group so the whole tree can be killed on
+    /// timeout.
+    fn spawn_inherited(
+        &self,
+        overlay: Option<&Path>,
+        subcommand: &[&str],
+    ) -> Result<ProcessGroupChild> {
+        let mut cmd: tokio::process::Command = self.command(overlay, subcommand).into();
```
carlsverre: why is this type assert here?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+                "failed to start '{} {}'",
+                self.cli.display(),
+                subcommand.join(" ")
+            )
+        })
+    }
+
+    /// Run `compose config [extra_args]`, returning the resolved YAML as a
+    /// string.
+    fn config_yaml(&self, overlay: Option<&Path>, extra_args: &[&str]) -> Result<String> {
```
carlsverre: Just call this config.

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+            .wrap_err_with(|| format!("failed to run '{name} ps'"))?;
+
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            return Err(eyre!("'{name} ps' failed"))
+                .with_section(move || stderr.trim().to_string().header("Stderr:"));
+        }
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        parse_compose_ps(&stdout)
```
carlsverre, line 402: Does ps support a structured output?

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+    }
+
+    /// Spawn `compose up --detach` and return the child process.
+    ///
+    /// stdout and stderr are inherited so progress is visible during pulls. The
+    /// caller awaits the child and checks its exit status. Uses
+    /// `process_group(0)` so the whole group can be killed on timeout.
+    pub fn up_detached(&self, overlay: Option<&Path>) -> Result<ProcessGroupChild> {
```
carlsverre, line 410: Rather than exposing the underlying command process group machinery to the caller, it would be cleaner to wrap the result in effectively a future. Analyze how this function is used and think about what semantic is actually needed. If the caller always just waits and gets output then we can just inline that. If the caller streams output somehow then we can expose that directly. Basically try to make the compose api feel like an api rather than exposing implementation details. This comment applies to all similar cases

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+/// content as building from `config_dir` directly.
+pub fn stage_pinned_config(config_dir: &Path, pinned_yaml: &str) -> Result<tempfile::TempDir> {
+    let staged = tempfile::tempdir().wrap_err("failed to create config staging directory")?;
+    copy_dir_recursive(config_dir, staged.path())?;
+    std::fs::write(staged.path().join("docker-compose.yaml"), pinned_yaml)
+        .wrap_err("failed to write pinned docker-compose.yaml")?;
+    Ok(staged)
+}
+/// Recursively copy the contents of `src` into `dst` (which must already exist).
+fn copy_dir_recursive(src: &Path, dst: &Path) -> Result<()> {
```
carlsverre: This is a utility. Is there no stdlib provider? If not is there an idiomatic crate for this kind of utility? If not move this to a utility mod.

src/compose.rs
```diff
@@ -0,0 +1,2057 @@
...
+{
+    let mut seen = HashSet::new();
+    let mut unsupported = Vec::new();
+
+    for image in images {
+        if !seen.insert(*image) {
+            continue;
+        }
+
+        let arch = runtime.image_architecture(image)?;
```
carlsverre, line 859: This fn should return an enum.

src/compose.rs
carlsverre: Done — inlined; the literal only appears in `resolve()` now. (071ba99)

src/compose.rs
carlsverre: Done — `compose_file_args` inlined into `command()`, and its two tests removed. (071ba99)

src/compose.rs
carlsverre: Done — `impl Display for ComposeCli`; hints and error context use it directly. (071ba99)

src/compose.rs
carlsverre: Done — removed the `restore_discovery_env` method and inlined the DOCKER_CONFIG restore into `config_json_hermetic_env` (its only caller), right next to the `env_clear()` it compensates for. (071ba99)

On resolving the plugin *path* instead (your other note): the docker CLI exposes no stable, machine-readable way to get the compose plugin binary path, so invoking it directly would mean reimplementing docker's cli-plugins discovery (`$DOCKER_CONFIG/cli-plugins` + system dirs) ourselves — more code and more fragile than the one-line env restore. Happy to revisit if you know a clean way to get the path.

src/compose.rs
carlsverre: Done — now checks an actual version number. `version()` runs `<cli> version --short` (e.g. `2.40.3`) and `require_v2()` parses the major component and requires `>= 2`. Dropped the `contains("docker compose")` product-name match. (071ba99)

src/compose.rs
carlsverre: Done — split the responsibilities. `version()` is now pure detection (returns the `--short` version string, no v2 judgment); the new `require_v2()` is where the caller (resolution) decides it needs v2+. (071ba99)

src/compose.rs
carlsverre: Done — renamed `config_yaml` → `config`. (071ba99)

src/compose.rs
carlsverre: It's not an assert — the annotation was only there to drive the `From<std::process::Command>` conversion for `.into()`. Made it explicit instead: `tokio::process::Command::from(self.command(...))`. (071ba99)

src/compose.rs
carlsverre: Deliberate layering, and `ComposeCli` is a private impl detail — `DockerCompose` is the only public type. `ComposeCli` answers "which compose form, and is it v2?" and is resolvable *without a project* — that's what lets `probe()` / `has_compose()` (doctor, tests) check availability with no config or engine. `DockerCompose` is a driver *bound to* a config dir + engine.

Merging them would force `config`/`docker_host` to be `Option` on the one type (probe has neither), reintroducing representable-but-invalid states — and the two forms still want the enum you asked for earlier. So it's really "enum for the two forms" + "struct that binds a form to a project", which is inherently two types. Open to merging if you still prefer it, but that's the reasoning.

src/compose.rs
carlsverre: Reasoned about both:

- **Current:** try standalone `docker-compose` (v2) first, else the `docker compose` plugin (guarding podman-in-disguise). Engine-agnostic — needs no `rt`, which is why `probe()`/`has_compose()` can resolve compose with no engine.
- **Engine-tied:** if the engine is genuine docker, use the plugin. Still needs the standalone fallback for podman, still needs the podman-in-disguise guard, *and* couples compose resolution to engine detection + requires threading `rt` into it (breaking the engine-free `probe` path). It also silently flips behavior for a docker user who has both installed (plugin instead of standalone).

I don't think it comes out simpler — it trades a `which` probe for an engine-conditional plus tighter coupling. Kept the current form; let me know if you disagree.

src/compose.rs
carlsverre: Yes — we already use it. The command is `ps -a --format json` and `parse_compose_ps` parses that JSON (handling both the NDJSON and array shapes compose has emitted across versions) into `Vec<ComposeContainer>`. There's no typed/stable Rust struct compose guarantees, so we deserialize the documented JSON.

src/compose.rs
carlsverre: Looked at how both callers use it (validate.rs `validate_compose`): neither "just waits for output". Each spawns the child and then **races** it in a `tokio::select!` — `up` against the `--timeout` deadline and ctrl-c; `logs --follow` against `watch_for_setup_complete` — and on the losing branch calls `kill_group()` to tear down the whole process tree (compose + its children). So the semantic genuinely needed is "a cancellable handle I can await *and* kill", not "await and get output".

`ProcessGroupChild` is already that abstraction (a snouty type in `process.rs`, not a raw tokio `Child`): it exposes `wait()` (awaitable) and `kill_group()`. Collapsing it into a plain future would drop the kill-on-cancel capability the callers rely on. I think this is the right level, but if you want I can wrap it in a compose-named handle to hide the "process group" wording.

src/compose.rs
carlsverre: Checked: `std::fs` has no recursive copy. The obvious crates don't fit — `fs_extra`/`walkdir`-based copies dereference symlinks, whereas we must **recreate symlinks verbatim** so the staged tree matches what `docker build` tars from the original context (that's the whole reason this exists). So no drop-in.

It's ~24 lines, single-use (only `stage_pinned_config`), and tied to that build-context semantic, so I kept it local rather than spinning up a `util` module for one function. Happy to extract it to a `util`/`fs` mod if you'd prefer it out of `compose.rs`.

src/compose.rs
carlsverre: Agreed that stringly-typed `"amd64"` is weak. It's the `ContainerRuntime::image_architecture` trait method in `container.rs` (returns `Result<String>`), consumed here by `validate_image_architectures` and `find_remote_pin`. Turning it into an `Architecture` enum touches the trait + all three impls (podman/docker/fake) + both call sites — meaningful, but it's container-runtime surface rather than the compose work this PR is about, and the PR is already large. I'd prefer to do it as a focused follow-up PR. Flagging it so it doesn't get lost.

carlsverre:
Addressed the latest round (commit `071ba99`) and replied inline on each thread.

**Implemented (8):**
- `impl Display for ComposeCli` (dropped `display() -> &str`)
- Version detection reworked: `version()` is now pure (`version --short`), and a new `require_v2()` parses the **major version number** and gates — replacing the product-name string match
- Inlined the single-use `DOCKER_COMPOSE` const, `compose_file_args` (into `command()`), and the `DOCKER_CONFIG` plugin-discovery restore (into its only caller, dropping the odd `restore_discovery_env` method)
- Renamed `config_yaml` → `config`
- `tokio::process::Command::from(...)` instead of the type-annotated `.into()`

**Responded with reasoning (7)** — kept as-is with rationale on the thread: why `ComposeCli`/`DockerCompose` are two types (private form-resolver vs project-bound driver), why resolution stays engine-agnostic, `ps` already uses `--format json`, the `up`/`logs` handle needs cancellable-wait + `kill_group` (both callers race it in `select!`), `copy_dir_recursive` has no symlink-preserving std/crate equivalent, and `image_architecture → enum` proposed as a focused follow-up (it's `ContainerRuntime` trait surface).

423 tests pass, clippy clean, `snouty doctor` verified.

src/compose.rs
carlsverre: Done — extracted to a new `util` module (`crate::util::copy_dir_recursive`). Kept the local impl since std has no recursive copy and the symlink-preserving requirement rules out the common crates, but it's no longer sitting in `compose.rs`. (77e9029)

src/compose.rs
carlsverre: Done in this PR — `ContainerRuntime::image_architecture` now returns an `Architecture` enum (`Amd64` | `Other(String)`) instead of a `String`. Callers compare against `Architecture::Amd64`; `Other` carries the raw name so error messages still name the offending arch (via `Display`). (77e9029)

carlsverre:
@claude review

src/doctor.rs
```diff
@@ -248,11 +248,16 @@ fn collect_checks(settings: &Settings) -> Vec<Check> {
...
+    // which one was picked.
+    match container::ComposeCommand::resolve().and_then(|c| c.version().map(|v| (c, v))) {
+        Ok((cmd, version)) => checks.push(Check::ok(
+            "docker_compose",
+            format!("{}: {version}", cmd.display()),
+        )),
         Err(e) => checks.push(
-            Check::fail("docker_compose", "docker-compose not available")
+            Check::fail("docker_compose", "Docker Compose v2 not available")
                 .note(Level::Error, e.to_string()),
```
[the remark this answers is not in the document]
carlsverre: having resolve return a version is codesmell. however as resolve computes the version, it's not unreasonable to just store the computed version in the ComposeCli type and then have version return that instead. lets do that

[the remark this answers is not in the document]
carlsverre: Done (commit `5356789`). `ComposeCli` is now a struct `{ form, version }`: `resolve()` runs `version --short` once via `ComposeForm::detect_v2()` (which both confirms v2 and returns the banner), stores it, and `version()` is a cheap `&str` accessor — no second spawn, and `resolve()` still returns just `Result<ComposeCli>` (no tuple). Verified with a call-counting wrapper: `snouty doctor` now spawns `compose version --short` exactly once (was twice).

src/compose.rs
```diff
@@ -0,0 +1,1969 @@
...
+/// The docker CLI config directory (which holds `cli-plugins/`): `$DOCKER_CONFIG`
+/// if set, else `$HOME/.docker`. Read from snouty's own (un-scrubbed) environment.
+/// `None` when neither is set, in which case plugin lookup falls back to the
+/// system directories.
+fn docker_config_dir() -> Option<PathBuf> {
+    if let Some(dir) = std::env::var_os("DOCKER_CONFIG") {
+        return Some(PathBuf::from(dir));
+    }
+    std::env::var_os("HOME").map(|home| PathBuf::from(home).join(".docker"))
+}
```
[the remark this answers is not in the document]
carlsverre: Done (commit `33543c7`). Split the pure core into `docker_config_dir_from(docker_config, home)` so it's unit-testable without mutating the process env, and added `docker_config_dir_prefers_explicit_then_home` covering all three branches (explicit DOCKER_CONFIG / $HOME/.docker / neither).

src/compose.rs
```diff
@@ -0,0 +1,1969 @@
...
+            .map(|h| format!("DOCKER_HOST={h} "))
+            .unwrap_or_default();
+        let compose_file = self.config.dir().join("docker-compose.yaml");
+        let mut hint = format!("{host}{} -f {}", self.cli, compose_file.display());
+        if let Some(overlay) = overlay {
+            hint.push_str(&format!(" -f {}", overlay.display()));
+        }
+        hint.push_str(" down");
+        hint
+    }
```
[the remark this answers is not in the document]
carlsverre: Good catch — fixed (commit `33543c7`). `down_hint()` now absolutizes both the compose-file path and the overlay via `std::path::absolute` (lexical, no I/O; falls back to the original on error), so the printed `--keep-running` teardown works from any directory even when `snouty validate` was given a relative path.

src/compose.rs
```diff
@@ -0,0 +1,1969 @@
...
+        let program = which::which("docker").wrap_err("`docker` not found on PATH")?;
+        // podman-in-disguise routes `docker compose` to a provider that may not
+        // implement Compose v2; never trust it as a v2 source.
+        if is_podman_in_disguise("docker") {
+            bail!("`docker` is podman in disguise; its `compose` is not Docker Compose v2");
+        }
+        let candidate = ComposeCli::Plugin(program);
+        candidate.require_v2()?;
+        Ok(candidate)
+    }
```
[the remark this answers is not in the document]
carlsverre: Fair point. Context: `is_podman_in_disguise` is pre-existing shared engine-detection code — `runtime()` and `available_engines()` already call it (running `docker version`) on essentially every invocation, independent of this PR; this PR adds one more call site (the compose plugin fallback). The unbounded `docker version` is the one that can block on a wedged daemon; the compose `version --short` probes are client-side.

Wrapping `is_podman_in_disguise` in `output_with_timeout(cmd, DISCOVERY_COMMAND_TIMEOUT)` is a clean, small robustness win and would cover all call sites consistently. It was raised and deferred in the earlier review round (as shared, pre-existing behavior). I've left it out of this PR to keep the scope on compose resolution — happy to fold in the timeout wrap (all call sites) here if you'd prefer, or track it as a follow-up.

src/compose.rs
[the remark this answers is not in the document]
carlsverre: Reverted (commit `ca8cea1`) — agreed the `docker_config_dir_from` split + test added more indirection than it was worth for a trivial `$DOCKER_CONFIG` else `$HOME/.docker` fallback, so it's back to the direct form.

Also relaxed AGENTS.md: unit tests for small/simple internal functions (trivial env/string/path plumbing) are now a judgement call rather than a blanket requirement.

PR #173, opened 2026-07-20
login: success confirmation, satisfaction gallery, and a real macOS keychain test

carlsverre:
Based on #157 (targets `jeskew/use_keychain`, not `main`, so it stacks on the login work rather than pushing to that branch while it's in flight).

### 1. `snouty login` confirms what it did (was silent on success)
Previously a successful `login` printed its last prompt and exited `0` with no confirmation — the user couldn't tell it worked, what changed, or the next step. The spec tests pass because they assert the *files*, not the human-facing feedback. It now prints:

```
Saved tenant `acme` and repository `registry.example.com/acme/app` to ~/.config/snouty/settings.toml.
Stored your API key in ~/.config/snouty/credentials.toml.
Run `snouty doctor` to verify your setup.
```

- Names the credential type and **where** it landed (system keychain vs `credentials.toml`), or that credential setup was skipped (env vars will be used).
- Names the **profile scope** when one is in use (`…under profile \`prod\`…`).
- Omits a blank repository (which is intentionally not persisted).

To report where credentials landed, `persist` now returns a `CredentialStorage` (`Keychain` | `File(path)`) and `update_settings_in_global_file` returns the path it wrote — both have a single caller.

### 2. `snouty login` folded into the satisfaction review + gallery
`scripts/gen-gallery.py` gains an interactive/stateful story kind: scripted stdin, a throwaway `$HOME`, seeded pre-existing state, and capture of the files login wrote (secrets redacted). Adds 8 `login-*` stories. They need no API, discovery, or container runtime, so `--only 'login-*'` skips discovery. The satisfaction-review skill documents the new transcript/persisted-state shape, reinterprets axes A–C for interactive commands, adds **Axis D (safety of a mutating command)**, and raises the width target to 140 cols.

### 3. A real macOS Keychain integration test
Every existing test forces `SNOUTY_DISABLE_KEYCHAIN_CREDENTIAL_STORAGE=1`, so the Security-framework path had zero coverage. `tests/keychain_macos.rs` drives it end-to-end on GitHub macOS runners: it creates a throwaway keychain, points the user default at it (snouty resolves `SecKeychain::default_for_domain(User)`, which honors `security default-keychain -d user -s`), runs `login`, verifies the credential is in the keychain and **not** a file, checks read-back and profile scoping, and restores + deletes the keychain via a `Drop` guard. It compiles everywhere (Linux CI type-checks it) and no-ops unless it's a macOS GitHub runner. No workflow change needed — the `macos-15-intel` matrix entry already runs `cargo nextest run`.

## Testing
`cargo fmt --check`, `cargo clippy --all-targets`, and the full test suite pass on Linux. The macOS keychain test only executes on the macOS runner (this PR's CI is where it first runs for real).

### DIFF
--- a/.claude/skills/satisfaction-review/SKILL.md
+++ b/.claude/skills/satisfaction-review/SKILL.md
@@ -71,6 +71,31 @@ _Automated check: PASS|FAIL — <detail>_
 (Help stories show an `Exit code:` line under each command block — the `--help`,
 the default output, and every variant.)
 
+**Interactive / stateful stories** (the `login-*` slugs) have a different shape.
+`snouty login` reads answers from stdin and *writes files*, so instead of a
+single output block these stories show:
+
+```
+## Transcript
+```shell
+$ snouty login
+# stdin> acme                     ← the answers fed to each prompt, in order
+# stdin> registry.example.com/...
+<the prompt transcript snouty printed>
+```
+Exit code: `<n>`
+
+## Persisted state
+`~/.config/snouty/settings.toml`      ← the files login actually wrote,
+```toml                                  with secrets already [REDACTED];
+...                                      "(not written)" when a file was
+```                                      deliberately not created
+```
+
+Judge these on the transcript **and** the persisted state together (see Step 3).
+They run in a throwaway `$HOME`, so the paths are temp dirs — judge the shape and
+content, not the exact path.
+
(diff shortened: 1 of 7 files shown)


jeskew (approved):
LGTM. Unfortunately, the macOS test won't run on a PR unless it's targeting `main`, so that may need some iteration once this PR lands

PR #176, opened 2026-07-21
Events v2: update OpenAPI spec and add --limit to runs events

carlsverre:
Adopts the events search API from the latest Antithesis release (orbitinghail tenant, release 58.6). Two commits, reviewable independently.

## 1. Update OpenAPI spec to tenant 58.6 and adapt launch client

Refreshed `src/openapi.json` from the tenant. The `/api/v0` and `/api/v1` spec routes now serve one identical combined document, and the tenant ships fully lenient schemas (no `additionalProperties: false` anywhere).

Client changes the new spec forced:

- **`build.rs`** — the schema-stripping transform is now a no-op. Per the assertion's own guidance it flips to **asserting the spec is free of `additionalProperties: false`** (reporting offending JSON-pointer paths), so a future spec that reintroduces it fails the build loudly rather than silently regenerating a brittle client.
- **Launch errors** — the launch endpoints now report failures with the new `Launch_Error_Response` envelope (`{statusCode, runId?}`) instead of the `{message}` shape. Routed through a shared, body-type-generic error formatter.
- **`run_id` is now nullable** on `LaunchResponse`/`LaunchMvdResponse`, and the documented debug-launch 202 body has converged with the live webhook envelope (`{runId, statusCode}`) — parsed directly now, carrying the body's own status code through. A now-redundant test (whose "spec vs live envelope differ" premise no longer holds) was removed.

Tests and specs updated to the `{statusCode, runId: null}` launch-error envelope, validated against the live tenant (an unknown launcher 404s with exactly that shape; gateway auth rejections carry an empty body instead, also covered).

The bulk of the diff is the regenerated `openapi.json`.

## 2. Add `--limit` to `snouty runs events`

The events endpoint now takes a `limit` query param (1–999) capping how many events the server returns.

- Exposed as **`-n/--limit`**, **unset by default**: with the flag omitted the query param is left off the request entirely, so tenants that predate it never receive it and the server applies its own default (currently 50). Range checking is left to the server.
- Raising the cap makes multi-needle searches (only the longest needle is matched server-side) more exhaustive, so the capped-subset note now points at `--limit`.
- The in-process mock honors the parameter (first N matches) so the behavior is covered end to end.

## Backwards compatibility

Tracked in [this comment](https://github.com/antithesishq/snouty/pull/176#issuecomment-5047783181) and resolved as follows:

- **`snouty debug` on older tenants** — resolved, no change needed: the API team confirmed `statusCode` has always been returned and the old spec was simply misaligned, so the recovery path this depended on can't fail. A missing `runId` is tolerated end to end.
- **Always sending `limit`** — fixed: the parameter is only sent when the user asks for one.
- **Older tenants' `{message}` launch errors** — still degrades to the generic invalid-payload path (the human-readable message is lost) when a pre-58.6 tenant returns `{message}` on a launch failure. Known remaining gap.

## Testing

`cargo test`, `cargo clippy --all-targets`, and `cargo fmt --check` all clean. Commit 1 also builds/tests in isolation.

### DIFF
--- a/specs/debug.txt
+++ b/specs/debug.txt
@@ -113,12 +113,13 @@ stdin moment_run_input.txt
 snouty --json debug --stdin
 stdout '\s*"runId": "run-123".*'
 
-# On API failure, the command exits with an error showing the HTTP status.
-# Launch endpoints report failures with the Launch_Error_Response envelope
-# ({ statusCode }), which carries no message text.
-mock-server 401 '{"statusCode":401}'
+# Gateway-level rejections (auth, rate limiting) on the launch/debug endpoints
+# return an EMPTY body rather than the Launch_Error_Response envelope — verified
+# against the live API, which answers a bad token with 401/403 and no payload.
+# snouty must surface that permissively instead of choking on the empty body.
+mock-server 401 ''
 ! snouty debug --input-hash abc123 --run-id 9043254f65c9c65d63fe043a0abfc7fc-53-1 --vtime 1234567890
-stderr 'API error: 401.*'
+stderr '.*empty response body.*'
 
 # ANTITHESIS_EXTRA_HEADERS adds arbitrary headers to every API request (for
 #    consumption by proxies). With --verbose, the merged request headers are
--- a/specs/debug.txt
+++ b/specs/debug.txt
@@ -113,9 +113,10 @@ stdin moment_run_input.txt
 snouty --json debug --stdin
 stdout '\s*"runId": "run-123".*'
 
-# On API failure, the command exits with an error showing the HTTP status
-#    and response body.
-mock-server 401 '{"message": "unauthorized"}'
+# On API failure, the command exits with an error showing the HTTP status.
+# Launch endpoints report failures with the Launch_Error_Response envelope
+# ({ statusCode }), which carries no message text.
+mock-server 401 '{"statusCode":401}'
 ! snouty debug --input-hash abc123 --run-id 9043254f65c9c65d63fe043a0abfc7fc-53-1 --vtime 1234567890
 stderr 'API error: 401.*'
 
--- a/specs/launch.txt
+++ b/specs/launch.txt
@@ -113,10 +113,11 @@ mock-server 200 '{"runId":"run-123","statusCode":202}'
 snouty launch -w basic_test --duration 30 --param 'antithesis.images=app@sha256:abc;db:latest'
 stderr '"antithesis.images": "app@sha256:abc;db:latest"'
 
-# On API failure, the command exits with an error showing the HTTP status.
-# Launch endpoints report failures with the Launch_Error_Response envelope
-# ({ statusCode }), which carries no message text.
-mock-server 400 '{"statusCode":400}'
+# On API failure that reaches the test launcher, the command exits with an
+# error showing the HTTP status. Such errors carry the Launch_Error_Response
+# envelope ({ statusCode, runId: null }), verified against the live API — there
+# is no message text, and runId is null once the request reaches the launcher.
+mock-server 400 '{"statusCode":400,"runId":null}'
 ! snouty launch -w basic_test --duration 30
 stderr 'API error: 400.*'
 
--- a/specs/launch.txt
+++ b/specs/launch.txt
@@ -113,13 +113,15 @@ mock-server 200 '{"runId":"run-123","statusCode":202}'
 snouty launch -w basic_test --duration 30 --param 'antithesis.images=app@sha256:abc;db:latest'
 stderr '"antithesis.images": "app@sha256:abc;db:latest"'
 
-# On API failure that reaches the test launcher, the command exits with an
-# error showing the HTTP status. Such errors carry the Launch_Error_Response
-# envelope ({ statusCode, runId: null }), verified against the live API — there
-# is no message text, and runId is null once the request reaches the launcher.
-mock-server 400 '{"statusCode":400,"runId":null}'
-! snouty launch -w basic_test --duration 30
-stderr 'API error: 400.*'
+# On an error that reaches the test launcher (e.g. an unknown launcher name),
+# the command exits with an error showing the HTTP status. Such errors carry
+# the Launch_Error_Response envelope ({ statusCode, runId: null }), verified
+# against the live API — no message text, and runId is null. (The launcher does
+# no request validation beyond auth, so it never returns a 400; 404 is the
+# canonical reachable failure.)
+mock-server 404 '{"statusCode":404,"runId":null}'
+! snouty launch -w no_such_launcher --duration 30
+stderr 'API error: 404.*'
 
 # `snouty run` still works but prints a deprecation warning.
 mock-server 200 '{"runId":"run-123","statusCode":202}'
--- a/specs/launch.txt
+++ b/specs/launch.txt
@@ -113,9 +113,10 @@ mock-server 200 '{"runId":"run-123","statusCode":202}'
 snouty launch -w basic_test --duration 30 --param 'antithesis.images=app@sha256:abc;db:latest'
 stderr '"antithesis.images": "app@sha256:abc;db:latest"'
 
-# On API failure, the command exits with an error showing the HTTP status
-#    and response body.
-mock-server 400 '{"message":"bad request"}'
+# On API failure, the command exits with an error showing the HTTP status.
+# Launch endpoints report failures with the Launch_Error_Response envelope
+# ({ statusCode }), which carries no message text.
+mock-server 400 '{"statusCode":400}'
 ! snouty launch -w basic_test --duration 30
 stderr 'API error: 400.*'
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -392,15 +392,21 @@ impl AntithesisApi {
         })
     }
 
-    pub async fn search_run_events(&self, run_id: &str, query: &str) -> Result<ByteStream> {
-        match self
+    pub async fn search_run_events(
+        &self,
+        run_id: &str,
+        query: &str,
+        limit: u16,
+    ) -> Result<ByteStream> {
+        // The endpoint caps the returned events at `limit`; the CLI always
+        // supplies one (defaulting to 50), and the server validates the range.
+        let request = self
             .client
             .search_run_events()
             .run_id(run_id)
             .q(query)
-            .send()
-            .await
-        {
+            .limit(u64::from(limit));
+        match request.send().await {
             Ok(response) => Ok(response.into_inner()),
             Err(err) => Err(format_api_client_error(err).await),
         }
@@ -1932,7 +1938,7 @@ mod tests {
         let api = test_api_optionally_with_cache(&mock_server, None);
 
         let mut stream = api
-            .search_run_events("run-1", "slow request")
+            .search_run_events("run-1", "slow request", 50)
             .await
             .unwrap()
             .into_inner();
@@ -1943,6 +1949,23 @@ mod tests {
         let body = String::from_utf8(body).unwrap();
 
         assert!(body.contains("slow request"));
+    }
+
+    #[tokio::test]
+    async fn search_run_events_forwards_limit() {
+        let mock_server = MockServer::start().await;
+
+        Mock::given(method("GET"))
+            .and(path("/api/v0/runs/run-1/events"))
+            .and(query_param("q", "slow"))
+            .and(query_param("limit", "5"))
+            .respond_with(ResponseTemplate::new(200).set_body_string(""))
+            .expect(1)
+            .mount(&mock_server)
+            .await;
+
+        let api = test_api_optionally_with_cache(&mock_server, None);
+        api.search_run_events("run-1", "slow", 5).await.unwrap();
     }
 
     fn rid(version: u32) -> String {
--- a/src/api.rs
+++ b/src/api.rs
@@ -248,31 +248,30 @@ impl AntithesisApi {
     pub async fn launch_debugging(&self, params: &Params) -> Result<LaunchResponse> {
         let body = launch_mvd_request(params)?;
         match self.client.launch_mvd().body(body).send().await {
-            // Documented 202 whose body is the spec shape (`{ "run_id": ... }`):
-            // the generated client parsed it into `LaunchMvdResponse` for us.
-            // Normalize to the `runId`/`statusCode` envelope the live API uses
-            // (and that `snouty launch` emits) so `--json` output is consistent.
-            Ok(response) => Ok(LaunchResponse {
-                run_id: response.into_inner().run_id,
-                status_code: 202,
-            }),
             // Documented 202 whose body is the live `{ "runId", "statusCode" }`
-            // envelope: the generated client expected the spec's snake_case
-            // `run_id` and rejected the camelCase body as an invalid payload.
-            //
-            // Unlike `UnexpectedResponse` below, this variant carries no HTTP
-            // status (`ClientError::status()` is `None`), so we can't gate on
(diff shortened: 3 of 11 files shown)


specs/debug.txt
```diff
@@ -113,9 +113,10 @@ stdin moment_run_input.txt
 snouty --json debug --stdin
 stdout '\s*"runId": "run-123".*'
 
-# On API failure, the command exits with an error showing the HTTP status
-#    and response body.
-mock-server 401 '{"message": "unauthorized"}'
+# On API failure, the command exits with an error showing the HTTP status.
+# Launch endpoints report failures with the Launch_Error_Response envelope
+# ({ statusCode }), which carries no message text.
+mock-server 401 '{"statusCode":401}'
```
carlsverre: message was removed? validate that reality against the real api.

specs/launch.txt
```diff
@@ -113,9 +113,10 @@ mock-server 200 '{"runId":"run-123","statusCode":202}'
 snouty launch -w basic_test --duration 30 --param 'antithesis.images=app@sha256:abc;db:latest'
 stderr '"antithesis.images": "app@sha256:abc;db:latest"'
 
-# On API failure, the command exits with an error showing the HTTP status
-#    and response body.
-mock-server 400 '{"message":"bad request"}'
+# On API failure, the command exits with an error showing the HTTP status.
+# Launch endpoints report failures with the Launch_Error_Response envelope
+# ({ statusCode }), which carries no message text.
+mock-server 400 '{"statusCode":400}'
```
carlsverre: validate against real api

src/api.rs
```diff
@@ -393,15 +392,21 @@ impl AntithesisApi {
         })
     }
 
-    pub async fn search_run_events(&self, run_id: &str, query: &str) -> Result<ByteStream> {
-        match self
+    pub async fn search_run_events(
+        &self,
+        run_id: &str,
+        query: &str,
+        limit: u16,
```
carlsverre: use a usize for limit everywhere. no need to artificially limit to 16k just to be "closer" to the real range. and the server may increase the limit in the future.

src/openapi.json
```diff
@@ -650,10 +666,23 @@
...
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "Maximum number of results to return. Must be between 1 and 999. Defaults to 50.",
+            "required": false,
+            "schema": {
+              "type": "integer",
+              "default": 50,
+              "minimum": 1,
+              "maximum": 999
```
carlsverre, line 685: does our openapi generator support generating a validation method? it would be cool if we could use the min/max values here to automatically validate outgoing requests locally. I'm ok with some validation that's tied directly to the openapi spec, but only if it's reasonably easy to do. Ideally the openapi generator is doing the heavy lifting. If it requires doing another pass over the openapi to figure out additional restrictions on our own I dont want to do it.

carlsverre (commented):
I talked to the API team about the changes that we are seeing in the spec as they may result in snouty being backwards incompatable with older tenants. Will hold on this PR until they get back to me on what the actual next release version of the openapi spec will look like. In the meantime we should flag all backwards incomptible changes so we are aware of the issues.

carlsverre:
## Backwards-incompatibility flag (per the hold on this PR)

I diffed the tenant-58.6 spec against `main`'s and validated what I could read-only against the live orbitinghail API. Here are the changes that could make this snouty build behave worse against **older (pre-58.6) tenants**, worst-first. None of these are snouty bugs — they're consequences of adopting the new spec, and they're the reason to wait for the API team's final envelope.

### 🔴 1. `snouty debug` breaks against older tenants on the documented 202 path
`Launch_MVD_Response` changed from `required: [run_id]`, `{ run_id: string }` to `required: [statusCode]`, `{ runId?: string, statusCode: int }` — the field was renamed (`run_id`→`runId`), made optional, and **`statusCode` became newly required**.

- An older tenant answers a debug launch with `{ "run_id": "..." }` and **no `statusCode`**.
- If it uses the documented **HTTP 202**, the generated `LaunchMvdResponse` fails to deserialize (missing `statusCode`) → `InvalidResponsePayload`. That arm deliberately only recovers when the *body* self-reports a 2xx `statusCode` (it can't see the transport status), which the old body lacks → **a successful debug launch is reported as an error.**
- If it uses the undocumented **HTTP 200**, we recover fine (the `UnexpectedResponse` arm defaults the status). b6114e0 makes even a `runId`-less 200 body succeed.

So the break is specifically: **old tenant + documented 202 + no `statusCode` in body.** Not fixable without a compat shim (e.g. treat `InvalidResponsePayload` carrying a `run_id` as success). Worth deciding once the API team confirms whether `statusCode` will really be required.

### 🟠 2. Launch/debug error messages degrade against older tenants
The launch endpoints' error responses moved from `Error_Response` (`{ message }`, via `BadRequest`) to `Launch_Error_Response` (`{ statusCode }`, via `Legacy_BadRequest`).

- Older tenants return `{ "message": "..." }` on a launch 4xx/5xx. The new client expects `{ statusCode }`, so the body no longer parses → we fall to the generic "invalid API response payload" path and **the human-readable message is lost.**
- Read-only reality check on 58.6: a **gateway auth rejection returns an empty body + 403** (not `{statusCode}`, not `{message}`), while non-launch 404s still return `{ "message": ... }`. So the real launch-error envelope is genuinely uncertain and depends on where in the stack the error originates — flagged on the two spec-mock threads. I couldn't `POST` from the sandbox to nail it down.

### 🟡 3. `snouty runs events` now always sends `limit`
`q` became optional and a new `limit` query param was added; snouty now sends `limit=50` on **every** events request, including to tenants that predate the parameter. Most servers ignore unknown query params (low risk), but if an older tenant validates them strictly this would 400. Unverified against an old tenant.

### ✅ Not a compat risk
- `Launch_Response` (`snouty launch`) kept `required: [runId, statusCode]`; `runId` only became nullable. Old tenants' `{runId, statusCode}` still parse. `snouty launch` is unaffected.
- The `additionalProperties: false → true` sweep only loosens the *server's* validation; snouty's client is already lenient about unknown fields, so it's a non-issue for parsing older responses.

### Addressed in this PR round (b6114e0)
- `parse_debug_launch_body` tolerates a missing run id (fixes the benign HTTP-200 half of #1 and the bot's nit).
- `--limit` is now `usize` (no artificial 16k ceiling; server owns the range).
- Confirmed progenitor 0.14 can't cheaply validate the `limit` max locally (see that thread).

Items #1–#3 are left as-is pending the API team's decision on the real next-release envelope.

src/api.rs
```diff
@@ -248,31 +248,30 @@ impl AntithesisApi {
...
+            // otherwise.
             Err(ClientError::InvalidResponsePayload(body, source)) => {
                 match parse_debug_launch_body(&body) {
                     Ok((run_id, Some(status_code))) if (200..300).contains(&status_code) => {
                         Ok(LaunchResponse {
-                            run_id,
+                            run_id: Some(run_id),
                             status_code,
                         })
                     }
```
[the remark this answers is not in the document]
carlsverre: Fixed in b6114e0. `parse_debug_launch_body` now returns `Option<String>` for the run id (a missing `run_id`/`runId` yields `None` instead of a hard error), matching the generated `LaunchMvdResponse` deserializer that the documented-202 path relies on. The two fallback arms pass it through unchanged. Added `launch_debugging_accepts_200_without_run_id` (undocumented-200 body `{"statusCode":202}` → `Ok`, `run_id: None`) plus a parser unit test, and removed the now-obsolete `parse_debug_launch_body_errors_without_a_run_id`.

Worth noting this bug is the *benign* half of a broader backwards-compat problem I've flagged in a top-level comment: the same schema change (`statusCode` now required, `runId` optional) means a **pre-58.6 tenant** returning the old documented-202 body `{"run_id":"..."}` (no `statusCode`) still fails, because that path can't self-report a 2xx status. That one isn't fixable without a compat shim, so I left it for the API-team discussion.

specs/debug.txt
carlsverre: I couldn't fully validate against the real API: `POST`s to the launch endpoints are blocked by the sandbox here. What I *could* observe read-only on the orbitinghail (58.6) tenant:

- A gateway auth rejection returns an **empty body with HTTP 403** (no `message`, no `statusCode`) — matching the spec's own note that "some gateway-level rejections … may instead return an empty body."
- A non-launch `404` still returns `{"message":"Resource not found"}` (the `Error_Response` envelope is alive and well for the non-launch endpoints).

So the `{statusCode}`-only launch-error envelope is **not confirmed** for the 401 case — reality looks more like an empty body there. This mock line is really asserting the *new* spec's claim, which is one of the backwards-incompat items in my top-level comment. Could you run this to confirm the real shape?

```
! KEY=$(python3 -c "import tomllib;print(tomllib.load(open('$HOME/.config/snouty/credentials.toml','rb'))['default']['api_key'])"); curl -s -i -X POST -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' -d '{"params":{"antithesis.duration":"1"}}' https://orbitinghail.antithesis.com/api/v1/launch/nonexistent_launcher_xyz
```

(A bad launcher name should 404 without launching anything.) Happy to adjust the mock body to whatever the real API actually returns.

specs/launch.txt
carlsverre: Same as the debug.txt thread: I can't `POST` from the sandbox to confirm, and read-only probes show gateway auth rejections return an **empty body / 403** on the real 58.6 tenant while non-launch endpoints still use `{"message":...}`. This line asserts the new spec's `{statusCode}` launch-error envelope, which I've listed as a backwards-incompat concern up top. See the debug.txt reply for a ready-to-run `curl` to validate the real shape; I'll match the mock to whatever it returns.

src/api.rs
carlsverre: Done in b6114e0 — `limit` is `usize` end to end now (cli → `cmd_runs_events` → `search_run_events`, converted to `u64` only at the generated-client boundary). Dropped the "1-999" wording from the flag help too, so the client doesn't assert a range the server owns and may raise.

src/openapi.json
carlsverre: Checked: progenitor 0.14 does **not** give us this cheaply. For the `limit` query param it uses only `minimum: 1` — mapping the builder argument to `NonZeroU64`, which is why `--limit 0` fails locally — but it ignores `maximum: 999` entirely (999 shows up only in a generated doc comment) and emits no validation method. So enforcing the max would mean our own second pass over the spec to extract bounds, which is exactly the "additional pass" you said you don't want. Leaving range validation to the server as-is. If progenitor grows constraint validation in a later release we can revisit for free.

specs/debug.txt
carlsverre: Validated against the live tenant. Confirmed: an auth/gateway rejection returns an **empty body** (bad token → HTTP 403, `content-type: text/plain`, zero-length payload), *not* a `{statusCode}` envelope. So the old `{"statusCode":401}` fixture was unfaithful. Updated in 693124b to `mock-server 401 ''`, asserting snouty surfaces the empty body permissively (`empty response body`) rather than choking. Also retargeted the empty-body suggestion away from the misleading "older-version" hint toward the real cause (gateway auth/rate-limit rejections carry no payload).

specs/launch.txt
carlsverre: Validated against the live tenant. An error that **reaches the launcher** (unknown launcher name → HTTP 404) returns exactly `{"statusCode":404,"runId":null}` (`content-type: application/json`) — confirming the `Launch_Error_Response` envelope, with `runId` present-but-null. Updated the fixture in 693124b to `{"statusCode":400,"runId":null}` to match that real shape. (Gateway rejections are the other case — empty body — now covered in specs/debug.txt.)

carlsverre:
## Live-API validation of the launch/debug error envelopes

Validated against the orbitinghail (58.6) tenant. Summary of what the launch endpoints actually return:

| Case | HTTP | Body | `content-type` |
|---|---|---|---|
| Error reaches the launcher (unknown launcher) | 404 | `{"statusCode":404,"runId":null}` | application/json |
| Gateway rejection (bad auth) | 403 | *empty* | text/plain |
| Missing / empty request body | 202 | `{"statusCode":202,"runId":"…"}` | application/json |

Two takeaways, both now reflected in the branch (693124b):

1. **`Launch_Error_Response` (`{statusCode, runId:null}`) is real** for launcher-reached errors, and snouty parses it into a clean `API error: <code>`.
2. **Gateway rejections carry no body**, and there is **no synchronous request validation** — a missing/invalid body launches with defaults and returns `202` rather than a 4xx. So snouty has to be permissive with launch/debug output: tolerate a missing `runId` on success (done), and an empty body on error (done — surfaced via the empty-body path, with the suggestion retargeted to auth/rate-limiting).

Fixtures updated: `specs/launch.txt` and the `launch_reports_api_errors` test now use the real `{statusCode, runId:null}` envelope; `specs/debug.txt` now covers the empty-body gateway case.

This resolves the "validate against the real API" review threads. The broader backwards-incompat items (#1 `snouty debug` on old-tenant 202 bodies, #2 old-tenant `{message}` launch errors) from the earlier flag still stand for the API-team discussion.

specs/launch.txt
carlsverre: Per your note that the launcher does no request validation beyond auth (so it never returns a 400), I retargeted this fixture from the never-occurring 400 to the **404 unknown-launcher** case — the canonical reachable launcher-level failure I actually confirmed against the live API — keeping the same `{statusCode, runId:null}` envelope. Done in b529d5b (same change applied to the `launch_reports_api_errors` test).

carlsverre:
1. we need to not send limit by default - as older tenants may not support the flag.
2. I talked to the API team and they confirmed that statusCode has always been returned, the issue was a misalignment with the spec. the run id may or may not be returned depending on tenant version + the possibility of a launch error

carlsverre:
Both points addressed.

### 1. `limit` is no longer sent by default (4ad6ec0, 4bfb3ce)

`--limit` is now `Option<usize>` end to end (`cli.rs` → `cmd_runs_events` → `search_run_events`), and the generated builder's `.limit()` is only called when a value was passed. With the flag omitted, the query parameter is left off the request entirely, so tenants that predate it never see it and the server applies its own default.

Coverage:
- `search_run_events_omits_limit_when_unset` — asserts `query_param_is_missing("limit")` against the mock.
- `search_run_events_forwards_limit` — unchanged behavior when `Some(5)` is passed.
- `events_limit_is_unset_by_default_and_parses` — clap yields `None` without the flag.
- `specs/runs.txt` — the no-`--limit` case now documents that no cap is sent at all (the in-process mock already treated `limit` as optional).

Flag help no longer asserts a client-side default: *"Maximum number of events the server returns (default 50)."*

### 2. `statusCode` is always returned — retires backwards-incompat item #1

Given the API team's confirmation that `statusCode` has always been on the wire and the old spec was simply misaligned, **item #1 from my earlier flag is resolved with no code change**. That item assumed a pre-58.6 tenant could answer a debug launch with a documented 202 carrying no `statusCode`, which would have failed the body's self-reported-2xx recovery gate in `launch_debugging`. Since `statusCode` is always present, that path can't trigger, and no compat shim is needed.

The "run id may or may not be returned" half is already tolerated end to end: `LaunchResponse.run_id` is `Option<String>`, `parse_debug_launch_body` returns `Option` for the id (both snake and camel case), and `snouty debug` prints `(unknown)` when it's absent — covered by `launch_debugging_accepts_200_without_run_id`.

That leaves item #3 (always sending `limit`) fixed here, item #1 resolved as above, and item #2 (older tenants' `{message}` launch errors degrading to the generic invalid-payload path) as the only remaining known compat gap.

`cargo test`, `cargo clippy --all-targets`, and `cargo fmt --check` are clean.

PR #179, opened 2026-07-28
runs logs: add --failure to stream logs at the run's failure moment

ValentaTomas:
`runs show` already resolves a run's failure moment and prints it as a ready-to-run `runs logs <run> <hash> <vtime>` command. This folds that into one step: `runs logs <run> --failure` fetches the run, resolves `real_failure_moment()`, and streams up to it — so the common "show me the logs at the failure" case needs no copy-paste of the moment.

- `INPUT_HASH`/`VTIME` become optional (`required_unless_present = "failure"`) and conflict with `--failure`.
- A run with no real failure moment gets a clear error pointing at `runs show`.
- Reuses the existing `real_failure_moment()` (so the placeholder 0/0 moment is treated as "none", same as `runs show`).

Tests: added a parse test for `--failure`, updated the two existing `logs` parse tests for the now-`Option` positionals; `cargo test --lib` green (427 passed). Left as a draft for your CI/clippy/formatting to confirm — my local run was in a container without your full toolchain.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -556,19 +556,31 @@ INPUT_HASH and VTIME identify the moment and its timeline; logs are streamed up
 to that moment. Without --begin-vtime, streaming starts at the timeline's
 earliest log entry.
 
+Pass --failure instead of HASH/VTIME to stream up to the run's own failure
+moment (the one `runs show` reports) — a shortcut for the common case of
+"show me the logs at the failure" without copying the moment by hand.
+
 Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
 `runs properties --detail` or `runs events`."#)]
     Logs {
         /// Run ID
         run_id: String,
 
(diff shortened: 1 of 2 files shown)


carlsverre:
Interesting idea! I'm a bit hesitant to do it though as it special cases the logs command in a way that is tied to a runs concept which may change in the future. Did you consider a simple wrapping script around `snouty runs show --json` to get the failure moment and then `snouty runs logs` to get the log? I feel like this is more suited to a script than a dedicated flag on the logs command. Does that make sense?

ValentaTomas:
@carlsverre Yes, that makes sense, I'm wrapping it already, but wanted to note this down in case it was interesting for merging.

carlsverre:
Appreciate it! Don't shy away from doing the same in the future. Also this is going to go into a "snouty cookbook" as a quick pattern which you just inspired me (claude) to build.

PR #183, opened 2026-07-28
Classify API failures on the HTTP status alone

### LINKED ISSUE #180: Launch responses carry a body-level statusCode alongside the HTTP status — collapse into one, and make error bodies permissive/status-preserving
## Open question for the API team

The launch endpoints (`POST /api/v1/launch/{launcher_name}`, `POST /api/v1/launch/debugging`) return a **body-level `statusCode`** in addition to the HTTP status of the response itself:

```
HTTP/1.1 202 Accepted
{"statusCode":202,"runId":"…"}
```

**Why does this field exist, and is it ever expected to diverge from the HTTP status?** The working hypothesis is that the API gateway supplies the transport status while the body's `statusCode` comes from the underlying API handler behind it — in which case the two can disagree whenever the gateway answers on its own behalf (auth rejection, rate limiting) rather than relaying the handler.

This issue is **blocked on the API team's answer**. Filing it so the question and the proposal don't get lost.

### What we know today

Divergence between the two is real, but every observed case is 2xx-vs-2xx:

- Nearly every launch fixture in `specs/launch.txt` is `mock-server 200 '{"runId":"run-123","statusCode":202}'` — HTTP 200 with a body that says 202. `src/api.rs` has a dedicated test for that undocumented-200 case (`launch_debugging_accepts_200_without_run_id`), and the mismatch is the sole reason `launch_debugging` carries both a documented-202 arm and an undocumented-2xx arm.
- Read-only probes against a 58.6 tenant, on the other hand, agreed in every case: unknown launcher → HTTP 404 + `{"statusCode":404,"runId":null}`; missing/empty request body → HTTP 202 + `{"statusCode":202,"runId":"…"}`; bad token → HTTP 403 with an **empty body** and `content-type: text/plain` (a gateway rejection — no envelope at all).
- There is no observed case of a 2xx HTTP status wrapping an error `statusCode`, or the reverse. We cannot currently prove it doesn't happen.

### Why this hurts snouty

1. **Success/failure classification depends on the body.** `AntithesisApi::launch_debugging` (`src/api.rs:248-303`) recovers a response whose body failed schema validation *only* when the body itself self-reports a 2xx `statusCode`. That gate exists because `progenitor_client::Error::InvalidResponsePayload` carries no HTTP status (see below), so the body was the only status available on that path.
2. **The generated client discards the HTTP status whenever an error body doesn't match its schema.** `Error::InvalidResponsePayload(Bytes, serde_json::Error)` has nowhere to put the status and `Error::status()` returns `None` for it (progenitor-client 0.14, `progenitor_client.rs:364`). Consequence, confirmed by test: `GET /api/version` answering **401 with an empty body** — exactly the shape the gateway returns for a bad token — is classified as `VersionError::Unreachable` rather than `Http(401)`, so `snouty doctor` reports *"Antithesis API unreachable / could not connect to \<host\>"* instead of *"Antithesis API rejected authentication"*. The same applies to any documented error status (401/406/429/500) whose body isn't a valid `Error_Response`, e.g. an HTML error page from an intermediary.
3. **Error-envelope churn is load-bearing.** Because error bodies are *typed* by the generated client, a change to the documented error schema (`Error_Response {message}` → `Launch_Error_Response {statusCode, runId}` in tenant 58.6) changes whether an error parses at all, and an unparseable error body downgrades a precise "HTTP 404 — \<message\>" into "invalid API response payload".

### Proposal

1. **Classify exclusively on the HTTP status.** The transport status decides success vs failure; the body is used only to enrich the message shown to the user (run id, any `message` text, raw payload snippet for debugging). This removes the `InvalidResponsePayload`-recovery arm and its "body self-reports 2xx" gate from `launch_debugging` entirely.
2. **Collapse the two status codes into one** on the API side, if the API team agrees the body-level field carries no information the HTTP status doesn't — i.e. stop returning it, or document precisely when it may differ so clients know which to trust.
3. **Make error responses permissive and status-preserving in the client.** Stop letting the generated client type error bodies at all: strip 4xx/5xx responses from the vendored spec in `build.rs` so every HTTP error surfaces as `UnexpectedResponse`, which retains both status *and* raw body, and funnel all of them through one status-first formatter (`format_api_error(status, body)` already does this). Then an empty, HTML, or unexpected-shape error body can never mask the status again, and the `Error_Response` / `Launch_Error_Response` split disappears.
4. **Interim safety net, while (2) is unresolved:** if the HTTP status is 2xx but the body's `statusCode` is >= 400, treat the launch as failed rather than silently reporting success. Drop this once the API team confirms the gateway cannot do that.
5. Consider filing upstream against progenitor so `InvalidResponsePayload` carries the response status — worth doing, but (3) means we no longer depend on it.

### Related

- The tenant-58.6 spec adoption that surfaced this: #176.
- The API team has confirmed that `statusCode` has **always** been present in launch responses and that the pre-58.6 OpenAPI simply failed to document it; `runId` may or may not be present depending on tenant version and whether the launch errored.

carlsverre:
Closes #180: make the HTTP status the only authority on whether an API call succeeded, and stop letting a non-conforming error body destroy it.

## What was broken

`progenitor_client::Error::InvalidResponsePayload(Bytes, serde_json::Error)` structurally has nowhere to put the HTTP status, and `Error::status()` returns `None` for that variant ([progenitor-client 0.14, `progenitor_client.rs:364`](https://docs.rs/progenitor-client/0.14.0/src/progenitor_client/progenitor_client.rs.html)). The generated client hits it whenever a **documented** status arrives with a body that doesn't match its schema — which is exactly the case we care about, because gateways and intermediaries answer with whatever they like:

- **`GET /api/version` → 401 with an empty body** — precisely what the API gateway returns for a bad token (`content-type: text/plain`, zero length) — was classified as `VersionError::Unreachable`, so `snouty doctor` printed *"Antithesis API unreachable / could not connect to \<host\>"* instead of *"Antithesis API rejected authentication / verify ANTITHESIS_API_KEY"*. Same for 406/429/500 behind an intermediary's HTML error page. This regressed when the 58.6 spec refresh (#176) gave `getVersion` typed error responses; v0.6.1 classified it correctly.
- **`launch_debugging`** recovered an unparsable body *only when the body itself self-reported a 2xx `statusCode`*, because with the transport status gone the body was the only status left. That gate also did `status_code.unwrap_or(202)`.
- Error-envelope churn was load-bearing: because error bodies are *typed*, an unparseable one downgraded a precise "HTTP 404 — \<message\>" into "invalid API response payload".

## The one detail the fix turns on

An **undocumented** status doesn't take that variant. It takes `Error::UnexpectedResponse(reqwest::Response)`, which keeps the status *and* the raw body. That is why a 404 on `getVersion` always worked correctly — it simply isn't documented for that operation.

So this bug is entirely a function of **which responses the spec tells progenitor to type**. `build.rs` already preprocesses the vendored spec, so that is where the fix goes.

## Approach: stop documenting error responses

One new transform, `untype_error_responses`, applied to the spec before generation: drop every non-2xx response from every operation.

Error bodies are precisely the ones that don't honour the schema — the gateway's empty `text/plain`, an intermediary's HTML page, an envelope the spec never described. Typing them let any of those shapes mask the status. Untyped, every failure reaches snouty as status + raw body, and one status-first formatter renders all of them. Dropping `default` responses along with the explicit 4xx/5xx matters too: progenitor turns a `default` into a typed catch-all covering every error status, which would reintroduce the problem by another door.

**Success responses are left alone.** The spec describes them accurately and the typed bodies are what the rest of the client is built on.

`src/openapi.json` is unmodified, and `build.rs`'s `additionalProperties: false` assertion is untouched.

### What that buys in `src/api.rs`

Every operation's error type becomes `Error<()>` — the generated client went from 64 `Error::ErrorResponse` arms to zero. `classify_client_error` collapses onto a single arm that always has both the status and the body exactly as the server sent it, and the `Error_Response` / `Launch_Error_Response` split disappears from the error path. There is now exactly one formatter (`format_api_client_error`) for every endpoint, and one `ApiFailure::into_report` stating how a failure renders.

`ClientError::ErrorResponse` is unreachable — progenitor only emits that arm for a documented error response — and `build.rs` asserts zero such arms in the generated source, in the same spirit as the existing `patch_lenient_booleans` and `additionalProperties` assertions. A spec refresh that reintroduces one fails the build. (Verified by disabling the transform: the build panics.) The runtime arm still reports the status rather than panicking, because progenitor *does* generate it for a documented error response with **no content** while leaving the error type `()` — so the type system alone doesn't rule it out, and a CLI shouldn't abort with a backtrace on something a server said.

Body text reaches the user through the existing status-first `format_api_error(status, body)`: unwrapped from `{"message": …}` when that's the shape, otherwise verbatim with whitespace collapsed and anything over 200 chars truncated, then run through `render::sanitize` — the repo's control-character policy — so an intermediary's error page stays a usable one-line diagnostic and can't smuggle escape sequences into the terminal.

`launch_debugging` loses its three-arm match, its body-status gate, its `unwrap_or(202)`, and `parse_debug_launch_body`; both launch entry points now share one `finish_launch` that treats any 2xx as success.

Authentication (the `pre` hook), verbose `format_request`/`format_response` logging, and the HTTP cache path in `send_request` are all untouched. Nothing buffers a response body, so the streaming NDJSON endpoints (`get_run_logs`, `get_run_build_logs`, `search_run_events`) are unaffected by construction. `specs/runs.txt` exercises all three and passes.

## Error bodies are bounded and sanitized

`read_error_body` reads at most 4 KiB of a failed response, chunk by chunk, and drops the rest — a misbehaving proxy shouldn't get snouty to buffer an arbitrarily large page to print 200 characters of it. What's shown is then collapsed to one line, run through `render::sanitize` (the repo's policy for untrusted text on a terminal line, so an error page can't emit escape sequences), and cut at 200 characters.

## `--json` no longer emits `statusCode` ⚠️ breaking

`snouty launch --json` and `snouty debug --json` now print `{ "runId": … }` and nothing else.

The field duplicated the HTTP status of a response the caller never sees, the exit code has always been the success signal, and [the API team has confirmed](https://github.com/antithesishq/snouty/issues/180#issuecomment-5110568006) that clients should ignore the body-level `statusCode` entirely. `LaunchResponse` is now a snouty-owned struct rather than a re-exported generated type, so the `--json` contract is a deliberate choice rather than whatever the vendored schema happens to say. Both specs assert the field's absence to pin that down.

**The break to watch for:** anything using `test $(snouty launch --json | jq .statusCode) -lt 300` as a success check now reads `null` rather than failing loudly. The exit code was always the correct signal, but this is a real backwards-incompatible output change.

The HTTP status remains available with `--verbose`, which dumps the response exactly as it arrived.

## User-visible changes

| Situation | Before | After |
|---|---|---|
| `doctor`, empty-body 401 from the gateway | "Antithesis API unreachable / could not connect to \<host\>" | "Antithesis API rejected authentication / the API returned HTTP 401" |
| any command, empty error body | "API returned an empty response body where a JSON payload was expected" | "API error: 401 Unauthorized" — the status is the whole message |
| `debug`/`launch`, HTML or unexpected-shape error body | "invalid API response payload: …" | "API error: 502 Bad Gateway — \<body\>" |
| launch error carrying `{"message": …}` | status line only (the field has no home in `Launch_Error_Response`) | message text surfaced |
| error body containing control characters | printed raw to the terminal | escaped as `\xNN`, like every other API-derived text |
| `launch --json` / `debug --json` | `{ runId, statusCode }` | `{ runId }` |

## Test matrix

New unit tests in `src/api.rs`:

- `get_version_keeps_the_status_for_non_conforming_error_bodies` — 401 empty, 401 HTML, 406 unexpected-shape JSON, 429 empty, 500 HTML all yield `VersionError::Http(<status>)`, not `Unreachable`.
- `get_version_keeps_the_status_for_an_undocumented_404` / `get_version_reads_the_documented_body` — regression guards on the paths that already worked.
- `launch_test_reports_the_status_for_any_error_body` — 404 with an empty body, an HTML page, and the live `{"statusCode":404,"runId":null}` envelope all report 404 (asserted structurally via `api_error_status`, not by string-sniffing).
- `launch_test_surfaces_the_error_message_body` — 404 + `{"message":"bad request"}` renders "API error: 404 Not Found — bad request".
- `launch_debugging_ignores_the_body_status_code` — HTTP 200 + `{"runId":"x","statusCode":202}` succeeds, and the serialized `--json` envelope is `{"runId":"x"}` alone.
- `launch_debugging_rejects_success_body_on_error_status` — 400/403/500 carrying `{"statusCode":202}` all error. Regression guard for the removed "body self-reports a 2xx" gate.
- `format_api_client_error_reports_the_status_for_an_empty_body` — 401, 429 and 502 with no body each keep their status.
- `read_error_body_*` (2) — an oversized body reads exactly the 4 KiB cap; a short one comes back intact.
- `error_body_message_*` (3) — envelope unwrapping, HTML collapsing/truncation, control-character escaping.

Specs: `specs/debug.txt` (empty-body 401 now names the status; new HTML-body 502 case; `--json` asserts `statusCode` is gone), `specs/launch.txt` (same `--json` assertion, 404 wording), `specs/doctor.txt` (new empty-body 401 → "rejected authentication" case; it runs in both modes — against staging, a bad key provokes the real gateway 401), `specs/runs.txt` (new `[!staging]` empty-body 429 → the status survives a body carrying nothing).

`cargo fmt --check`, `cargo clippy --all-targets`, the full `cargo test` suite (433 lib + all integration and spec targets), and `typos` are clean.

## Revision history

**`7f9d1ba` — review round.** Dropped the empty-body suggestion (the status line already says everything the server said), bounded the error body read to 4 KiB, simplified `error_body_message` now that the bound makes a hand-rolled bounded collapse unnecessary, and put `ClientError::ErrorResponse` back to `unreachable!`.

**`2695223`** — collapsed the launch-specific formatter into the shared one (the note it carried was removed a round later), and converted the last hand-rolled launch mocks in the test module to the shared helper.

**`6697749` — quality pass.** Routed error body text through `render::sanitize`, which is the repo's control-character policy; `error_body_message` had grown a weaker private copy that only collapsed Unicode whitespace, so ESC/BEL from a proxy's error page reached the terminal unescaped. Also bounded the collapse (it used to rebuild an entire error page into a `Vec` and a `String` before discarding all but 200 chars), gave `ApiFailure` an `into_report` so the rendering rule is stated once, and folded three near-identical wiremock helpers into one.

**`e34986a` — review round.** Reverted the launch-success half of the spec transform, dropped `statusCode` from `--json`, and simplified `finish_launch`'s signature.

The reverted piece deserves calling out, because its justification was wrong. It claimed the launch body shape "varies by tenant release (`runId` vs `run_id`)". Backwards: the **pre-58.6 spec** declared `Launch_MVD_Response` with a required *snake_case* `run_id`, while the live webhook has always sent camelCase `runId`. That mismatch is what 9ec1f37 worked around and what #176 fixed at the source. No tenant emits `run_id`, so nothing needs to tolerate it — and `statusCode` is `required` in the 58.6 schema and always sent. The spec is accurate for launch success; the transform was inventing a compatibility requirement.

**`ac9352f` — approach change.** The first revision fixed this at runtime, in the progenitor `exec` hook: it buffered every error body, synthesized an envelope engineered to satisfy `Error_Response` and `Launch_Error_Response` simultaneously so the failure would land in `Error::ErrorResponse` with its status intact, and smuggled the real body past progenitor base64-encoded on a private `x-snouty-original-body` response header. Every one of those steps existed to *feed* a typed error path we don't want; documenting fewer responses removes the path instead.

That removed `normalize_response`, `normalized_launch_body`, `rebuild_response`, `original_body`, `is_launch_operation`, `ORIGINAL_BODY_HEADER`, and the `base64` use in `api.rs`.

Measured against `main`:

| | `src/api.rs` | `build.rs` | net new |
|---|---|---|---|
| `7ea9ea9` (`exec` hook) | +559 / −242 | — | **+317** |
| `7f9d1ba` (this) | +428 / −312 | +63 | **+179** |

Also considered and rejected:

- **Editing `src/openapi.json` directly.** It's vendored and refreshed wholesale (#176); the edit would vanish on the next refresh. `build.rs` is the established place, with precedent in both directions (`patch_lenient_booleans` today, the `additionalProperties` strip in git history).
- **A different generator.** openapi-generator's Rust target models errors as `ResponseContent<E>`, which does keep the status, so it would fix the bug — but it adds a Java toolchain to the build, rewrites every call site and every generated type name, and snouty's auth `pre` hook and cache-aware `exec` hook are progenitor-specific. Wildly disproportionate to one variant's missing field.
- **Configuring progenitor.** `GenerationSettings` in 0.14 has no error-handling knob; `with_patch` / `with_replacement` / `with_conversion` are typify type-level controls (derives, renames, substitute types). The only lever progenitor gives you here is what the spec documents.
- **Filing upstream** so `InvalidResponsePayload` carries the response status (issue proposal item 5) is still worth doing, but snouty no longer depends on it.

### Known trade-off

Untyping error responses gives up typed access to `Error_Response.message`. In practice `error_body_message`'s generic `{"message": …}` unwrap covers it, and the premise of #180 is that error bodies come from intermediaries that don't honour the schema anyway — but if the API later adds structured error fields (an error code, say), they'd be hand-parsed.

Rebased onto `main` after #190 merged, so the specs run on the forked testscript engine. Every pattern this PR adds carries a regex trigger character, so the engine change does not alter their meaning. All checks re-verified after the rebase, including staging against a live tenant.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -64,12 +64,10 @@ fn generate_api_client(out_dir: &Path) {
 /// — would otherwise only surface as a status-less failure against a live
 /// tenant.
 fn assert_no_typed_error_responses(content: &str) {
-    let arms = content.matches("Error::ErrorResponse").count();
-    assert_eq!(
-        arms, 0,
-        "generated client has {arms} `Error::ErrorResponse` arm(s); every error response is \
-         supposed to be undocumented so failures keep their HTTP status (see \
-         `untype_error_responses`)"
+    assert!(
+        !content.contains("Error::ErrorResponse"),
+        "generated client has an `Error::ErrorResponse` arm; every error response is supposed \
+         to be undocumented so failures keep their HTTP status (see `untype_error_responses`)"
     );
 }
 
--- a/build.rs
+++ b/build.rs
@@ -41,7 +41,7 @@ fn generate_api_client(out_dir: &Path) {
         offenders.len(),
         offenders.join(", ")
     );
-    relax_response_typing(&mut spec_value);
+    untype_error_responses(&mut spec_value);
     let spec: openapiv3::OpenAPI = serde_json::from_value(spec_value).unwrap();
 
     let mut settings = progenitor::GenerationSettings::default();
@@ -52,12 +52,29 @@ fn generate_api_client(out_dir: &Path) {
     let ast = syn::parse2(tokens).unwrap();
     let content = prettyplease::unparse(&ast);
     let content = patch_lenient_booleans(content);
+    assert_no_typed_error_responses(&content);
 
     fs::write(out_dir.join("antithesis_api.rs"), content).unwrap();
 }
 
-/// Stop the generated client from typing the response bodies the server is
-/// known to answer off-schema.
+/// Hold `untype_error_responses` to its promise: with no error response
+/// documented anywhere, progenitor emits no `Error::ErrorResponse` arm, so that
+/// variant is unreachable and `classify_client_error` treats it as such. A spec
+/// refresh that slipped one back in — under a status shape the transform misses
+/// — would otherwise only surface as a status-less failure against a live
+/// tenant.
+fn assert_no_typed_error_responses(content: &str) {
+    let arms = content.matches("Error::ErrorResponse").count();
+    assert_eq!(
+        arms, 0,
+        "generated client has {arms} `Error::ErrorResponse` arm(s); every error response is \
+         supposed to be undocumented so failures keep their HTTP status (see \
+         `untype_error_responses`)"
+    );
+}
+
+/// Stop the generated client from typing error response bodies, so an HTTP
+/// failure always reaches snouty with its status attached.
 ///
 /// progenitor decodes every *documented* response into a generated type. When a
 /// documented status arrives with a body that doesn't match its schema, the
@@ -65,40 +82,22 @@ fn generate_api_client(out_dir: &Path) {
 /// serde_json::Error)` — a variant with nowhere to put the HTTP status, for
 /// which `Error::status()` is `None`. An *undocumented* status takes
 /// `Error::UnexpectedResponse(reqwest::Response)` instead, which keeps the
-/// status *and* the raw body. Two families of response are better left
-/// undocumented, so snouty always has the status:
+/// status *and* the raw body.
 ///
-/// - **Every error response, on every operation.** An error body is whatever
-///   the thing that produced it feels like sending: the API gateway rejects a
-///   bad token with an empty `text/plain` body, an intermediary answers with an
-///   HTML page. Typed, any of those masks the status — an empty-bodied 401 was
-///   reaching `snouty doctor` as "API unreachable" (#180). Untyped, every
-///   failure arrives as status + raw body and one status-first formatter
-///   renders all of them.
+/// Error bodies are exactly the ones that don't honour the schema: the API
+/// gateway rejects a bad token with an empty `text/plain` body, an intermediary
+/// answers with an HTML page. Typed, any of those masked the status — an
+/// empty-bodied 401 was reaching `snouty doctor` as "API unreachable" (#180).
+/// Untyped, every failure arrives as status + raw body and one status-first
+/// formatter renders all of them.
 ///
-/// - **The launch webhooks' success body.** The spec documents 202 while the
-///   live webhook answers 200, and the body shape varies by tenant release
-///   (`runId` vs `run_id`; `statusCode` present or not). Their success response
-///   is collapsed into a single `2XX` range carrying a free-form object, so any
-///   2xx is a success and `finish_launch` reads the body itself. Keeping a
-///   documented success — rather than dropping the response outright — is what
-///   preserves the generated `Accept: application/json` request header, which
-///   these endpoints content-negotiate on (they document a 406).
-fn relax_response_typing(spec: &mut serde_json::Value) {
-    const LAUNCH_OPERATIONS: [&str; 2] = ["launchTest", "launchMvd"];
-
-    let launch_success = serde_json::json!({
-        "2XX": {
-            "description": "Any 2xx is a successful launch. The body shape varies by tenant release, so it is left free-form and parsed by the client.",
-            "content": { "application/json": { "schema": { "type": "object" } } }
-        }
-    });
-
+/// Success responses are left alone: the spec describes them accurately, and
+/// the typed bodies are what the rest of the client is built on.
+fn untype_error_responses(spec: &mut serde_json::Value) {
     let paths = spec
         .get_mut("paths")
         .and_then(serde_json::Value::as_object_mut)
         .expect("openapi spec has a `paths` object");
-    let mut launch_operations_found = 0;
     for path_item in paths.values_mut() {
         let Some(path_item) = path_item.as_object_mut() else {
             continue;
@@ -106,36 +105,18 @@ fn relax_response_typing(spec: &mut serde_json::Value) {
         // A path item holds non-operation keys too (`parameters`, `summary`);
         // an operation is anything that documents responses.
         for operation in path_item.values_mut() {
-            let is_launch = operation
-                .get("operationId")
-                .and_then(serde_json::Value::as_str)
-                .is_some_and(|id| LAUNCH_OPERATIONS.contains(&id));
             let Some(responses) = operation
                 .get_mut("responses")
                 .and_then(serde_json::Value::as_object_mut)
             else {
                 continue;
             };
-            if is_launch {
-                launch_operations_found += 1;
-                *responses = launch_success.as_object().unwrap().clone();
-            } else {
-                // Also drops any `default` response, which progenitor would
-                // otherwise turn into a typed catch-all for every error status.
-                responses.retain(|status, _| status.starts_with('2'));
-            }
+            // Retaining only 2xx also drops any `default` response, which
+            // progenitor would otherwise turn into a typed catch-all covering
+            // every error status — reintroducing the problem by another door.
+            responses.retain(|status, _| status.starts_with('2'));
         }
     }
-
-    // A spec refresh that renamed or dropped a launch operation would otherwise
-    // silently leave its responses typed, and the mismatch would only show up
-    // as a runtime "invalid API response payload" against a live tenant.
-    assert_eq!(
-        launch_operations_found,
-        LAUNCH_OPERATIONS.len(),
-        "expected the openapi spec to define {LAUNCH_OPERATIONS:?}; found {launch_operations_found} \
-         of them, so the launch response bodies may still be typed"
-    );
 }
 
 /// Recursively collect the JSON-pointer path of every `"additionalProperties":
--- a/build.rs
+++ b/build.rs
@@ -18,7 +18,7 @@ fn main() {
 
 fn generate_api_client(out_dir: &Path) {
     let file = std::fs::File::open("src/openapi.json").unwrap();
-    let spec_value: serde_json::Value = serde_json::from_reader(file).unwrap();
+    let mut spec_value: serde_json::Value = serde_json::from_reader(file).unwrap();
 
     // A schema's `additionalProperties: false` makes progenitor/typify emit
     // `#[serde(deny_unknown_fields)]`, which turns a forwards-compatible server
@@ -41,6 +41,7 @@ fn generate_api_client(out_dir: &Path) {
         offenders.len(),
         offenders.join(", ")
     );
+    relax_response_typing(&mut spec_value);
     let spec: openapiv3::OpenAPI = serde_json::from_value(spec_value).unwrap();
 
     let mut settings = progenitor::GenerationSettings::default();
@@ -53,6 +54,88 @@ fn generate_api_client(out_dir: &Path) {
     let content = patch_lenient_booleans(content);
 
     fs::write(out_dir.join("antithesis_api.rs"), content).unwrap();
+}
+
+/// Stop the generated client from typing the response bodies the server is
+/// known to answer off-schema.
+///
+/// progenitor decodes every *documented* response into a generated type. When a
+/// documented status arrives with a body that doesn't match its schema, the
+/// generated client returns `Error::InvalidResponsePayload(Bytes,
+/// serde_json::Error)` — a variant with nowhere to put the HTTP status, for
+/// which `Error::status()` is `None`. An *undocumented* status takes
+/// `Error::UnexpectedResponse(reqwest::Response)` instead, which keeps the
+/// status *and* the raw body. Two families of response are better left
+/// undocumented, so snouty always has the status:
+///
+/// - **Every error response, on every operation.** An error body is whatever
+///   the thing that produced it feels like sending: the API gateway rejects a
+///   bad token with an empty `text/plain` body, an intermediary answers with an
+///   HTML page. Typed, any of those masks the status — an empty-bodied 401 was
+///   reaching `snouty doctor` as "API unreachable" (#180). Untyped, every
+///   failure arrives as status + raw body and one status-first formatter
+///   renders all of them.
+///
+/// - **The launch webhooks' success body.** The spec documents 202 while the
+///   live webhook answers 200, and the body shape varies by tenant release
+///   (`runId` vs `run_id`; `statusCode` present or not). Their success response
+///   is collapsed into a single `2XX` range carrying a free-form object, so any
+///   2xx is a success and `finish_launch` reads the body itself. Keeping a
+///   documented success — rather than dropping the response outright — is what
+///   preserves the generated `Accept: application/json` request header, which
+///   these endpoints content-negotiate on (they document a 406).
+fn relax_response_typing(spec: &mut serde_json::Value) {
+    const LAUNCH_OPERATIONS: [&str; 2] = ["launchTest", "launchMvd"];
+
+    let launch_success = serde_json::json!({
+        "2XX": {
+            "description": "Any 2xx is a successful launch. The body shape varies by tenant release, so it is left free-form and parsed by the client.",
+            "content": { "application/json": { "schema": { "type": "object" } } }
+        }
+    });
+
+    let paths = spec
+        .get_mut("paths")
+        .and_then(serde_json::Value::as_object_mut)
+        .expect("openapi spec has a `paths` object");
+    let mut launch_operations_found = 0;
+    for path_item in paths.values_mut() {
+        let Some(path_item) = path_item.as_object_mut() else {
+            continue;
+        };
+        // A path item holds non-operation keys too (`parameters`, `summary`);
+        // an operation is anything that documents responses.
+        for operation in path_item.values_mut() {
+            let is_launch = operation
+                .get("operationId")
+                .and_then(serde_json::Value::as_str)
+                .is_some_and(|id| LAUNCH_OPERATIONS.contains(&id));
+            let Some(responses) = operation
+                .get_mut("responses")
+                .and_then(serde_json::Value::as_object_mut)
+            else {
+                continue;
+            };
+            if is_launch {
+                launch_operations_found += 1;
+                *responses = launch_success.as_object().unwrap().clone();
+            } else {
+                // Also drops any `default` response, which progenitor would
+                // otherwise turn into a typed catch-all for every error status.
+                responses.retain(|status, _| status.starts_with('2'));
+            }
+        }
+    }
+
+    // A spec refresh that renamed or dropped a launch operation would otherwise
+    // silently leave its responses typed, and the mismatch would only show up
+    // as a runtime "invalid API response payload" against a live tenant.
+    assert_eq!(
+        launch_operations_found,
+        LAUNCH_OPERATIONS.len(),
+        "expected the openapi spec to define {LAUNCH_OPERATIONS:?}; found {launch_operations_found} \
+         of them, so the launch response bodies may still be typed"
+    );
 }
 
 /// Recursively collect the JSON-pointer path of every `"additionalProperties":
--- a/specs/doctor.txt
+++ b/specs/doctor.txt
@@ -37,11 +37,13 @@ stderr '.*snouty launch.*'
 stderr '.*ANTITHESIS_PASSWORD.*'
 
 # --- API connectivity + version check (issue #146) ---
-# Cases that inject a status via mock-server are [!staging]-gated: under
+# Cases that fabricate a status via mock-server are [!staging]-gated: under
 # SNOUTY_STAGING, mock-server passes through to the real API where fabricated
-# responses don't hold. The --offline and unreachable cases use a controlled
-# local setup and run in both modes. The mock + --offline cases assume a
-# container runtime (CI provisions one); the unreachable case is robust either way.
+# responses don't hold. The 401 case is the exception — a bad key provokes the
+# same status from the real gateway, so it runs in both modes. The --offline
+# and unreachable cases use a controlled local setup and run in both modes. The
+# mock + --offline cases assume a container runtime (CI provisions one); the
+# unreachable case is robust either way.
 
 # 2xx: doctor reports the live API and tenant release versions. The check runs
 # only with an API key (the endpoint rejects basic auth), so set one on top of
@@ -67,12 +69,15 @@ stderr '.*ANTITHESIS_PASSWORD.*'
 # token (content-type: text/plain, zero length). The HTTP status is what
 # classifies the failure, so doctor names the rejected credentials instead of
 # reporting the API as unreachable because the body didn't parse (issue #180).
-[!staging] mock-server 401 ''
-[!staging] env ANTITHESIS_API_KEY=testkey
-[!staging] ! snouty doctor
-[!staging] stderr '.*Antithesis API rejected authentication.*'
-[!staging] stderr '.*returned HTTP 401.*'
-[!staging] ! stderr '.*Antithesis API unreachable.*'
+# This block runs against staging too: mock-server passes through to the real
+# gateway, and the bad key below provokes the real 401, so the issue #180
+# scenario is exercised end to end rather than only against a mock of it.
+mock-server 401 ''
+env ANTITHESIS_API_KEY=testkey
+! snouty doctor
+stderr '.*Antithesis API rejected authentication.*'
+stderr '.*returned HTTP 401.*'
+! stderr '.*Antithesis API unreachable.*'
 
 # --offline skips the network check entirely: no version line, no API contact.
 env ANTITHESIS_API_KEY=testkey
--- a/specs/doctor.txt
+++ b/specs/doctor.txt
@@ -70,13 +70,16 @@ stderr '.*ANTITHESIS_PASSWORD.*'
 # classifies the failure, so doctor names the rejected credentials instead of
 # reporting the API as unreachable because the body didn't parse (issue #180).
 # This block runs against staging too: mock-server passes through to the real
-# gateway, and the bad key below provokes the real 401, so the issue #180
-# scenario is exercised end to end rather than only against a mock of it.
+# gateway, and the bad key below provokes a real rejection, so the issue #180
+# scenario is exercised end to end rather than only against a mock of it. The
+# mock answers 401; the live gateway answers 403 for a bad key (the probes on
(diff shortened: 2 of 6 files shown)


carlsverre:
Force-pushed a rewrite: `7ea9ea9` → `2e6e2e2`. The previous three commits are still reachable at `7ea9ea9` if you want to diff the two approaches directly.

Same behaviour, different place. The old revision normalized responses at runtime in the progenitor `exec` hook — buffering error bodies, synthesizing an envelope that satisfied both documented error schemas, and carrying the real body past the generated client base64-encoded on a private response header. The new one changes what the spec documents in `build.rs` instead, so the status-losing code path (`Error::InvalidResponsePayload`) is never reached rather than worked around.

Every behavioural test from the old revision passes **unmodified**, and all three spec files are byte-identical to the old revision. The five tests that are gone existed only to pin down the runtime machinery (`normalized_launch_body_*`, `launch_responses_require_a_status_code_field`).

src/api.rs
```diff
@@ -25,11 +25,26 @@ mod generated {
...
+/// snouty owns this shape rather than re-exporting a generated type: the launch
+/// webhooks' response body varies by tenant release, so `build.rs` leaves it
+/// free-form and [`finish_launch`] reads the run id out of it directly.
+/// `status_code` is the response's HTTP status, never the body's own copy of it
+/// (#180).
+#[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LaunchResponse {
+    pub run_id: Option<String>,
+    pub status_code: u16,
```
carlsverre: drop this, I don't want snouty to expose the status code in the output of launch or debug

carlsverre: (obviously I don't mean that we should hide the status code if verbose is specified, I'm talking about the normal output)

src/api.rs
```diff
@@ -784,56 +745,38 @@ fn launch_request(params: &Params) -> Result<generated::types::LaunchRequest> {
...
+/// live API uses (200) and the body's shape varies by tenant release. So any
+/// 2xx is a success, and the only thing read out of the body is the run id —
+/// under either spelling. The reported `statusCode` is the transport status:
+/// the API team has confirmed the body's own copy should be ignored (#180).
+async fn finish_launch(
     result: std::result::Result<
-        progenitor_client::ResponseValue<T>,
-        ClientError<generated::types::LaunchErrorResponse>,
+        progenitor_client::ResponseValue<serde_json::Map<String, serde_json::Value>>,
+        ClientError<()>,
```
carlsverre: simplify these fully qualified types with `use` if you can

src/api.rs
```diff
@@ -784,56 +745,38 @@ fn launch_request(params: &Params) -> Result<generated::types::LaunchRequest> {
...
-        .or_else(|| value.get("runId"))
-        .and_then(serde_json::Value::as_str)
-        .map(str::to_string);
-    let status_code = value.get("statusCode").and_then(serde_json::Value::as_i64);
-    Ok((run_id, status_code))
+/// The run id in a launch response body, under either spelling: tenants from
+/// 58.6 on answer with the documented `runId`, older ones with `run_id`. Absent
+/// on a launch that started no run, and on the debugging webhook's 200, which
+/// answers `{"statusCode": 202}` alone.
+fn launch_run_id(body: serde_json::Map<String, serde_json::Value>) -> Option<String> {
```
carlsverre: inline this

src/api.rs
```diff
@@ -1039,74 +982,120 @@ fn char_pos_to_byte_offset(body: &str, line: usize, column: usize) -> usize {
...
+                status,
+                message: error_body_message(body),
+            }
+        }
+        // No operation documents an error response, so the generated client has
+        // no typed error body to construct this from.
+        ClientError::ErrorResponse(response) => ApiFailure::Response {
+            status: response.status().as_u16(),
+            message: String::new(),
+        },
```
carlsverre: so this should be unreachable?

build.rs
```diff
@@ -55,6 +56,88 @@ fn generate_api_client(out_dir: &Path) {
...
+///   preserves the generated `Accept: application/json` request header, which
+///   these endpoints content-negotiate on (they document a 406).
+fn relax_response_typing(spec: &mut serde_json::Value) {
+    const LAUNCH_OPERATIONS: [&str; 2] = ["launchTest", "launchMvd"];
+
+    let launch_success = serde_json::json!({
+        "2XX": {
+            "description": "Any 2xx is a successful launch. The body shape varies by tenant release, so it is left free-form and parsed by the client.",
+            "content": { "application/json": { "schema": { "type": "object" } } }
+        }
```
carlsverre: why do we need to change launch responses? the openapi spec seems correct for launch responses already. I'm pretty sure it always returns 202 with optional runId and required statusCode. I think we are overstepping by messing with launch success in this PR.

src/api.rs
carlsverre: Done in `eb2e872`. `LaunchResponse` is now just `{ runId }`, and both specs assert `statusCode` is absent so the run-id-only envelope is pinned as a deliberate contract rather than drifting back in.

Worth flagging as the one backwards-incompatible bit in the PR: anything doing `test $(snouty launch --json | jq .statusCode) -lt 300` as a success check now reads `null` instead of breaking loudly. The exit code was always the correct signal, so I think that is the right trade, but it is a real break and I did not want it buried.

carlsverre: Right — `--verbose` is unaffected. `format_response` in the `exec` hook dumps the response exactly as it arrived, status line included, and this PR does not touch that path. I put a pointer to it in the `LaunchResponse` doc comment and in both spec comments so the next reader knows where the status went.

src/api.rs
carlsverre: Done. Imported `ResponseValue` from `progenitor_client`, and the launch-success revert (your `build.rs` comment) collapsed the body type back to the generic, so the signature is now:

```rust
async fn finish_launch<T: DeserializeOwned>(
    result: std::result::Result<ResponseValue<T>, ClientError<()>>,
) -> Result<T>
```

`std::result::Result` stays qualified because `Result` in this file is eyre's one-parameter alias — same as `get_version` already does.

src/api.rs
carlsverre: Gone entirely rather than inlined — `launch_run_id` only existed to read the free-form launch body, which the `build.rs` revert removes. The generated types deserialize `runId` directly again.

src/api.rs
carlsverre: Yes — and it is now written that way. progenitor only emits an `Error::ErrorResponse` arm for a *documented* error response, and after the transform there are none: the generated client went from 64 of those arms to 0.

I did not want to rely on that by inspection, though, since a spec refresh could reintroduce an error response under a status shape the transform misses and the only symptom would be a status-less failure against a live tenant. So `build.rs` now greps its own generated output and asserts zero `Error::ErrorResponse` arms before writing it, in the same spirit as the existing `patch_lenient_booleans` and `additionalProperties` assertions. With that guarantee in place the arm is a plain `unreachable!` pointing back at the transform. I verified the assertion actually fires by disabling the transform — the build panics rather than producing a client that compiles.

build.rs
carlsverre: You are right, and my justification for it was wrong — reverted in `eb2e872`.

I claimed the launch body shape "varies by tenant release (`runId` vs `run_id`)". That is backwards. Checking `22afaaf~1`, the **pre-58.6 spec** declared `Launch_MVD_Response` with a *required snake_case* `run_id`:

```json
"Launch_MVD_Response": { "required": ["run_id"], "properties": { "run_id": {...} } }
```

The live webhook has always sent camelCase `runId`. So the snake_case spelling was the spec being wrong about the server, never a server that emitted it — that is exactly what 9ec1f37 was working around, and #176 fixed it at the source. There is no tenant to tolerate. Same for "`statusCode` present or not": it is `required` in the 58.6 schema and the API team confirmed it is always sent.

So the spec is accurate for launch success and I was fabricating a compatibility requirement. `relax_response_typing` is now `untype_error_responses` and strips error responses only, which is the actual bug. `finish_launch` is back to its pre-existing shape, and the `2XX` range, `launch_run_id`, the snake_case spec case and its test are all gone. The `Accept: application/json` argument I used to justify keeping a documented success goes away with it — the real documented 202 carries that header on its own.

I have corrected the PR description too; it asserted the same wrong thing.

carlsverre:
Review addressed in `eb2e872` (a new commit, so `2e6e2e2..eb2e872` is the delta against what you read). Replies are on each thread; the short version:

- **`build.rs` launch responses** — you were right, and my justification for the change was factually wrong. The pre-58.6 *spec* declared a required snake_case `run_id` while the live webhook always sent `runId`; I had it as a tenant that emits snake_case. Reverted: the transform now strips error responses only, launch success stays typed by the spec, and `finish_launch` is back to its pre-existing shape.
- **`statusCode` in `--json`** — dropped from both `launch` and `debug`. Both specs now assert its absence. Flagging that this is the one backwards-incompatible change in the PR: a `jq .statusCode` success check now reads `null` rather than failing loudly.
- **`ErrorResponse` unreachable** — yes. Made it `unreachable!`, and added a `build.rs` assertion that the generated client contains zero `Error::ErrorResponse` arms so it can't rot back into reachability. Verified the assertion fires by disabling the transform.
- **Fully-qualified types / inline** — signature simplified with a `ResponseValue` import; `launch_run_id` is gone entirely with the revert.

The PR description asserted the same wrong thing about tenant versions and has been corrected, with a revision-history section at the bottom.

Net effect on the diff: `src/api.rs` +374/−263 and `build.rs` +65, down from +559/−242 in the first revision. `cargo test` (432 lib + all integration and spec targets), `clippy --all-targets`, `fmt --check`, and `typos` all clean.

carlsverre:
@claude review

src/api.rs
carlsverre: Follow-up in `257f068`: the three remaining hand-rolled launch mocks in the test module (`launch_test_sends_snouty_user_agent`, `launch_test_uses_basic_auth`, `launch_test_accepts_200_webhook_envelope`) now go through `mock_launch_test`, with the success envelope as a named `LAUNCH_OK_BODY` constant. That drops the last duplicate `Mock::given(...)` blocks — the module has one way to mock an endpoint now.

src/api.rs
carlsverre: Also in `257f068`, related to the `--json` contract: `format_launch_client_error` is gone and the empty-body note is attached by the shared formatter, so `snouty runs` against a rate-limited tenant explains its silence the way `snouty launch` already did instead of printing a bare `API error: 429 Too Many Requests`. `specs/runs.txt` gains that case.

src/api.rs
```diff
@@ -1039,74 +983,144 @@ fn char_pos_to_byte_offset(body: &str, line: usize, column: usize) -> usize {
...
+/// `content-type: text/plain` and zero length — and it is not specific to any
+/// endpoint: `GET /api/version` answers exactly the same way, which is what
+/// `snouty doctor` trips over in #180. The note deliberately says nothing about
+/// credentials, because [`format_api_error`] already attaches that on a 401/403
+/// and this fires for 429 and 5xx too.
+async fn format_api_client_error(err: ClientError<()>) -> Report {
+    match classify_client_error(err).await {
+        ApiFailure::Response { status, message } if message.is_empty() => {
+            format_api_error(status, "").with_suggestion(|| {
+                "the API answers some gateway-level rejections (for example authentication or rate limiting) with an empty body, so there is no detail to show beyond the status; if the problem persists, contact Antithesis support."
```
carlsverre: remove this suggestion. the error is sufficient.

src/api.rs
carlsverre: so this should now be unreachable!

src/api.rs
```diff
@@ -1039,74 +983,144 @@ fn char_pos_to_byte_offset(body: &str, line: usize, column: usize) -> usize {
...
+async fn classify_client_error(err: ClientError<()>) -> ApiFailure {
+    match err {
+        // Every HTTP failure lands here, carrying both its status and its body
+        // exactly as the server sent them.
         ClientError::UnexpectedResponse(response) => {
             let status = response.status().as_u16();
             let body = response
                 .text()
                 .await
                 .unwrap_or_else(|err| format!("<failed to read response body: {err}>"));
```
carlsverre: when reading the response body into a string we should put some reasonable bound on it. like 4KiB or something

src/api.rs
```diff
@@ -1039,74 +983,144 @@ fn char_pos_to_byte_offset(body: &str, line: usize, column: usize) -> usize {
...
+    let mut collapsed = String::new();
+    for word in body.split_whitespace() {
+        if collapsed.len() > MAX_LEN * 4 {
+            break;
+        }
+        if !collapsed.is_empty() {
+            collapsed.push(' ');
+        }
+        collapsed.push_str(word);
+    }
```
carlsverre: what value is this over just truncating the string to the first 200 chars and replacing newlines?

src/api.rs
carlsverre: Removed in `7c4227f`. The status line already says everything the server said, and `format_api_error` still attaches the credentials hint on a 401/403, so the paragraph was pure noise.

That also collapses the last of the formatter split: one match in `format_api_client_error`, and `ApiFailure::into_report` folds back into it. Dropped the two spec assertions that pinned the note; `specs/runs.txt` keeps the empty-bodied 429 case itself, since it is still the regression guard that the status survives a body carrying nothing.

src/api.rs
carlsverre: Done — 4 KiB, in `7c4227f`. `read_error_body` reads chunk by chunk and stops at the cap instead of `response.text()` buffering whatever a proxy chose to send. Truncating mid-character is fine here since the result is only ever displayed, and `from_utf8_lossy` renders a split character as U+FFFD.

Two tests: an oversized body reads exactly `MAX_ERROR_BODY`, and a short one comes back intact so the bound never eats a real error message.

src/api.rs
carlsverre: Reverted to `unreachable!` in `7c4227f`.

For the record on why I had softened it: progenitor generates that arm for a documented error response with *no content* while leaving the operation's error type as `()`, so the variant is not ruled out by the type system alone — only by the build-time assert. But that assert is doing its job, and a loud invariant beats a quietly-handled impossible case. Your call, and it is a better one.

src/api.rs
carlsverre: Fair — and with the 4 KiB read bound from your other comment, the elaborate version I had lost its reason to exist. It is now a plain `split_whitespace().collect::<Vec<_>>().join(" ")`; the bounded hand-rolled loop only existed to avoid rebuilding a multi-megabyte page, which cannot happen once the read is capped.

What collapsing buys over replacing newlines alone: an indented error page spends the 200 characters on its text rather than on its margins. `<html>\n      <body>Bad Gateway` costs ~8 characters of budget as margin per line if only the newline is replaced. On a deeply indented proxy page that is most of the window. Now documented on the function rather than left implicit.

The other half is `sanitize`, which is the repo's policy for untrusted text on one terminal line (`runs.rs` puts every API-derived cell through it) — it escapes ESC/BEL and friends so a proxy's error page cannot write escape sequences to the terminal. Running it after the collapse leaves it nothing to do about newlines and only control characters to escape.

specs/doctor.txt
```diff
@@ -63,6 +63,17 @@ stderr '.*ANTITHESIS_PASSWORD.*'
...
+# 401 with an EMPTY body — exactly what the API gateway returns for a bad
+# token (content-type: text/plain, zero length). The HTTP status is what
+# classifies the failure, so doctor names the rejected credentials instead of
+# reporting the API as unreachable because the body didn't parse (issue #180).
+[!staging] mock-server 401 ''
+[!staging] env ANTITHESIS_API_KEY=testkey
+[!staging] ! snouty doctor
+[!staging] stderr '.*Antithesis API rejected authentication.*'
+[!staging] stderr '.*returned HTTP 401.*'
+[!staging] ! stderr '.*Antithesis API unreachable.*'
```
carlsverre: can this be run against staging?

carlsverre: Yes. Under `SNOUTY_STAGING`, `mock-server` passes through to the real gateway, and the bad API key then provokes a real 401 with an empty body — the exact issue #180 scenario, end to end. The gates were the only obstacle: the fork pinned in #190 lets a condition on a negated line skip cleanly, and the assertions themselves hold against any tenant.

Done in `7c90974`: the block loses its `[!staging]` prefixes and now runs in both modes, so the `staging` CI job exercises it against the real gateway on every pull request. The section comment names the exception.

PR #184, opened 2026-07-29
Bound and retry the two flaky macOS podman CI steps

carlsverre:
Two unrelated-to-the-code failures cost three red CI runs on #181 and #183 over the last day, both on the `macos-15-intel` podman leg. Neither had anything to do with the changes under test; #181 (a 16-line comment-and-logging change in `src/auth.rs`) went red twice before passing on a third run.

## 1. A hung `podman machine start` presented as a bare cancelled job

`podman machine init`/`start` carry no timeout of their own. One run sat in `start` for **16m 34s** until the job's `timeout-minutes: 20` killed it — the `Test` step never ran, and the only signal was `##[error]The operation was canceled.` with a `Starting machine "podman-machine-default"` line above it. Nothing said provisioning had stalled.

Provisioning now runs as its own step with `timeout-minutes: 8` and `continue-on-error: true`, followed by a retry step gated on `steps.podman_machine.outcome == 'failure'` that discards the machine and rebuilds it. A wedged start is not sticky, so one retry recovers the leg instead of burning a whole run.

Splitting the retry into a second step is what makes it *bounded*: the first attempt's `timeout-minutes` has already been spent, and the retry gets its own budget. That is also why there's no shell-level timeout here — I tried `perl -e 'alarm shift; exec @ARGV'`, the usual dependency-free substitute for the coreutils `timeout` macOS lacks, and **it does not work on podman**: podman is a Go program and the Go runtime absorbs SIGALRM. Measured against a deliberately hung Go binary, the alarm never fired and an outer `timeout` had to do the killing. Step-level `timeout-minutes` is the only bound that reliably stops these commands, so that's what this uses. The script notes it so the idiom doesn't get reintroduced.

Sizing: healthy runs take 110–170s to init and 156–170s to start, so 8m is ~1.4× the healthy pair and less than half the observed hang. The job budget goes 20m → 35m because it now has to exceed the sum of the step bounds — otherwise the job timeout fires first and hides the attribution the step bounds exist to provide.

The provisioning commands moved into `scripts/provision-podman-machine.sh` so both attempts size the VM identically, matching how `scripts/build-validate-samples.sh` is already shared between CI and local use.

## 2. The registry image pull had no retry

`OCIRegistry::start` pulled `registry:2` exactly once and discarded both the reason and the output on failure. On the macOS runner that pull crosses the VM's network proxy; when it failed, every registry-backed test failed with it — two tests in the observed run — and the message was just `OCI registry image could not be pulled with podman`, which required digging through the job log to learn nothing more.

`ensure_registry_image_available` now retries three times with a 1s pause, and returns the runtime's own last line plus its exit status on failure, so the skip/panic message says *why*:

```
OCI registry image could not be pulled with podman: Error: copying system image: proxy connection refused (exit status: 125)
```

Retrying is close to free: after a success the image is local, so there is nothing to re-fetch. A spawn failure (the binary itself missing) is returned immediately rather than retried, since that will never recover.

## Testing

`cargo fmt --check`, `cargo clippy --all-targets`, and the full `cargo test` are clean, and I ran the suite three times over to check for interference from the new sleeps. New unit tests cover the retry count on success (exactly one pull), the retry-then-report path (all three attempts, stderr and exit status both surfaced), the non-retryable spawn failure, and the one-line truncation.

The workflow change is validated as far as it can be off-runner: YAML parses, the extracted script passes `bash -n`, and the retry gate uses stock `steps.<id>.outcome` semantics. The behaviour it targets only reproduces on the macOS runner, so this PR's own macOS run is the first real exercise of the happy path — the retry path stays unexercised until provisioning next wedges.

## Not addressed

The `Load or pull k8s-validator image (macOS)` step has the same un-retried-pull shape and its comment already documents that the implicit pull "wedges indefinitely" under parallel load. It's cached, so it fails far less often; worth the same treatment if it starts showing up.


.github/workflows/build.yml
```diff
@@ -65,9 +65,14 @@ jobs:
...
-    # macOS in well under 10m, so 20m is generous headroom.
-    timeout-minutes: 20
+    # timeout) must not hang the runner for hours. Linux finishes in ~80s;
+    # macOS is much slower, and its steps carry their own `timeout-minutes`
+    # (below) because a job-level timeout reports a bare cancellation with no
+    # clue which step stalled. This budget has to exceed the sum of those step
+    # bounds, or it fires first and hides the very attribution they exist for:
+    # install 8 + provision 8 + provision retry 8 + validator image ~2 + tests
+    # ~9 = ~35m.
+    timeout-minutes: 35
```
carlsverre: minimize comment, and can we use a matrix param for timeout? we don't need to increase all timeouts, just the macos one.

.github/workflows/build.yml
```diff
@@ -121,16 +126,10 @@ jobs:
       - name: Install Podman + docker-compose (macOS)
         if: matrix.runtime == 'podman' && startsWith(matrix.os, 'macos')
         shell: bash
+        # Downloads and installs; ~25s when healthy, so a stall here is a
+        # network problem worth reporting quickly rather than sitting on.
+        timeout-minutes: 8
```
carlsverre, line 127: remove comment

.github/workflows/build.yml
```diff
@@ -141,24 +140,48 @@ jobs:
...
-          t=$SECONDS
-          podman machine start
-          start_s=$(( SECONDS - t ))
+      # `podman machine init`/`start` carry no timeout of their own and do
+      # sometimes wedge on this runner: one observed run sat in `start` for 16m
+      # until the job timeout killed it, reporting only a cancelled job. Giving
+      # provisioning its own step with `timeout-minutes` bounds that hang and
+      # names it, and `continue-on-error` lets the retry below act on it.
+      # Healthy runs take 110-170s to init and 156-170s to start, so 8m is
+      # roughly 1.4x the healthy pair and well short of the observed hang.
```
carlsverre: minimize comment

.github/workflows/build.yml
```diff
@@ -141,24 +140,48 @@ jobs:
...
 
-          printf '[timing] brew install: %ss | machine init: %ss | machine start: %ss\n' \
-            "$brew_s" "$init_s" "$start_s"
-          echo "::notice title=podman setup timing::brew install ${brew_s}s | machine init ${init_s}s | machine start ${start_s}s"
+      # A wedged start is not sticky — discarding the machine and building a
+      # fresh one clears it — so retry once instead of failing the whole macOS
+      # leg on a transient hang. A second step is what makes the retry bounded
+      # too: the first attempt's `timeout-minutes` has already expired, and this
+      # one gets its own budget. Without `continue-on-error` here, a second
+      # failure fails the job, which is what we want.
```
carlsverre: minimize comment

scripts/provision-podman-machine.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Bounding a hang is the workflow's job, not this script's: each attempt runs as
+# its own step with `timeout-minutes`, which is the only bound that reliably
+# stops these commands. macOS ships no coreutils `timeout`, and the usual
+# dependency-free substitute — perl's `alarm`, which survives execve — does not
+# work here, because podman is a Go program and the Go runtime absorbs SIGALRM.
+#
+# Run it by hand the same way CI does:
+#
+#   scripts/provision-podman-machine.sh
```
carlsverre: minimize comment, remove any notes on workload/time/bounding/perl. just keep things simple

src/testutils.rs
```diff
@@ -824,14 +877,106 @@ mod tests {
...
+            "a pull that works must not be retried, got: {log}"
+        );
+    }
+
+    // A failing pull is retried, and the runtime's own complaint reaches the
+    // caller — the macOS registry flake used to surface as a bare "could not be
+    // pulled" with the reason discarded.
+    #[cfg(unix)]
+    #[test]
+    fn ensure_registry_image_available_retries_and_reports_the_failure() {
```
carlsverre: remove this test

src/testutils.rs
```diff
@@ -824,14 +877,106 @@ mod tests {
...
+                .count(),
+            REGISTRY_PULL_ATTEMPTS as usize,
+            "expected {REGISTRY_PULL_ATTEMPTS} attempts, got: {log}"
+        );
+    }
+
+    // A runtime binary that cannot be executed is not a transient failure, so it
+    // is reported without burning the retries.
+    #[test]
+    fn ensure_registry_image_available_does_not_retry_a_missing_runtime() {
```
carlsverre: remove this test

src/testutils.rs
```diff
@@ -824,14 +877,106 @@ mod tests {
...
+            .expect_err("a missing runtime binary must report an error");
+
+        assert!(
+            err.contains("could not run"),
+            "expected a spawn failure, got: {err}"
+        );
+    }
+
+    #[test]
+    fn describe_pull_failure_truncates_and_keeps_the_last_line() {
```
carlsverre: remove this test

.github/workflows/build.yml
carlsverre: Done — `timeout-minutes: ${{ matrix.timeout }}`, with 20 on both Linux legs and 35 only on macOS. Comment cut to three lines.

.github/workflows/build.yml
carlsverre: Removed.

.github/workflows/build.yml
carlsverre: Cut to three lines.

.github/workflows/build.yml
carlsverre: Cut to three lines.

scripts/provision-podman-machine.sh
carlsverre: Trimmed to three lines: what it does, and that CI runs it twice so both attempts size the VM the same. The perl/SIGALRM and timing notes are gone — the reasoning lives in the PR description if anyone needs it later.

src/testutils.rs
carlsverre: Removed.

src/testutils.rs
carlsverre: Removed.

src/testutils.rs
carlsverre: Removed.

PR #190, opened 2026-08-01
Repair the staging spec path and cover it in CI

carlsverre:
`SNOUTY_STAGING=1 cargo nextest run spec_tests` cannot pass. It stops on the third spec file, before it reaches the `runs` specs it exists for. This PR repairs that path, fixes two assertions that could never hold against a real tenant, and makes CI gate merges on the result.

## The bug that stopped the run

testscript-rs 0.2.10 splits command execution across two functions:

```rust
fn execute_command(env, command, params) -> Result<()> {
    let result = execute_command_inner(env, command, params); // returns Ok(()) when the condition skips the line
    if command.negated {
        match result {
            Ok(_) => Err(Error::command_error(&command.name,
                "Command was expected to fail but succeeded")),   // a skipped line lands here
            Err(_) => Ok(()),
        }
    } else { result }
}
```

The inner function checks the condition and returns `Ok(())` for a skipped line. The outer function reads that `Ok` as success. So a line that carries both a condition and a `!` never skips. It fails the file instead.

Three lines did this: `specs/proxy.txt:24`, `:25`, and `specs/runs.txt:39`. Under staging the first one stops the run:

```
Test 'specs/proxy.txt' failed: Error in specs/proxy.txt at line 24:
> 24 | [!staging] ! snouty runs
Command 'snouty' failed: Command was expected to fail but succeeded
```

A normal run meets the condition and executes the line, so it passes. Only the mode nobody runs was broken. 0.2.10 is the latest release.

## Two assertions that could never hold

The first run against a real tenant (`orbitinghail`, release 58.10) found two more bugs that the mock hid.

`runs.txt` asserted `stdout 'Run ID .*run-1'` with no prefix, in two blocks. `run-1` is mock data. A real tenant returns a run hash. The unprefixed line now checks the structural claim, and the mock run id moves behind `[!staging]`.

Writing that structural check exposed the second bug. testscript compiles a pattern as a regex only when the pattern contains one of `^ $ [ ( * .`, and compares any other pattern as literal text. `'Run ID\s+\S+'` has none of those characters, so `\s` and `\S` were literal and the pattern matched nothing, in both modes. It is now `'Run ID +[^ ]'`.

## The fixes

Both bugs are now fixed at the source, in the [antithesishq fork of testscript-rs](https://github.com/antithesishq/testscript-rs) (its PR #1). This PR pins the fork's `main`:

- A condition on a negated line skips the line instead of failing the file.
- `stdout` and `stderr` always read the pattern as a regex, as Go's testscript does. `-literal` reads it as plain text. Nothing guesses from the pattern's contents, so `'Run ID\s+\S+'` is a hard regex now rather than silently-inert literal text.
- The regex flags are Go's: `(?m)`, so `^` and `$` match at line boundaries and `.` stops at a newline. A pattern that spans lines states `(?s)`, and the two in `runs.txt` already did.
- `-count=0` and a negated `-count` are rejected, as Go rejects them. `! stdout 'x'` is the zero-match assertion.

Spec fallout is small. The six `stdout -count=0 'x'` workaround lines become `[!staging] ! stdout 'x'`, which the fixed engine handles. One assertion gains `-literal` (it matches the two characters `\n` inside a JSON string, which regex mode reads as a newline). `proxy.txt` is restructured onto one `[staging] skip` with the offline settings check above it; that check now sets its own credentials, because it was inheriting them from the blocks that used to run first.

The inert-pattern lint this PR carried in earlier revisions is gone. The fork removed the failure class it guarded: a pattern is always a regex now, so a regex escape with no `^ $ [ ( * .` character is valid instead of silently-inert literal text, and the lint had started rejecting correct specs. An invalid regex is the engine's own error.

Point the dependency back at crates.io when upstream ships these fixes.

## CI

The `staging` job in `build.yml` gates the merge. It runs beside the `test` matrix, takes 57s to 3m19s depending on the cargo cache, and the macOS leg takes 8 to 11 minutes. It does not change the total CI time.

The job reads `STAGING_ANTITHESIS_TENANT` and `STAGING_ANTITHESIS_API_KEY`, and sets no base URL, so snouty resolves `https://<tenant>.antithesis.com`. A missing secret fails the job, because a check that does not run is worse than a check that fails.

> [!NOTE]
> The job skips on a fork's pull request. GitHub does not pass secrets to a workflow that a fork's pull request triggers, and approving the run does not change this: "With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository."
>
> `tests-passed` accepts `skipped` for this job and no other result. To run the suite on a fork's branch, push it to this repo: `gh pr checkout <n> && git push origin HEAD:<branch>`.

## One behaviour change that gating forces

The suite must not depend on tenant data. One block filtered `--status incomplete` and asserted the result carries a failure moment. That holds on `orbitinghail` today. It fails on a tenant whose runs all passed, and `env_from_json` would capture nothing. That block moves behind `[!staging]`. The completed run covers the same endpoints.

One requirement remains, and AGENTS.md records it: **the tenant must have at least one completed run.**

## What the suite covers

| | |
|---|---|
| commands that run under staging | 147 of 201 |
| commands that reach the API | 31 |
| API operations covered | 6 of 9 |
| assertions in the reachable part of `runs.txt` | 68 total, 45 skipped, 23 run |

The value is not in the assertion count. Every API call must deserialize a real response through the generated client. If `src/openapi.json` drifts from what the API returns, the command fails whatever the assertion says. The mock cannot catch this, because its fixtures come from the spec. The suite also covers real authentication, cursor pagination, and release-version gating.

`launchTest` and `launchMvd` stay out, because they mutate tenant state. `getRunLogs` is read-only but sits below the `[staging] skip` in `runs.txt`, so it is out as well. Recovering it is follow-up work.

## Testing

`cargo test` (427 lib tests plus every integration and spec target), `clippy --all-targets`, `fmt --check`, and `typos` are clean, all against the pinned fork. The staging suite passes against `orbitinghail` with the same configuration CI uses. The fork itself is green on its own three-OS CI (Linux, macOS, Windows; 122 tests).

Branched off `main`. Independent of #183.


.github/workflows/build.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches: ["main"]
   pull_request:
     branches: ["main"]
+  # Nightly run on main. The `staging` job is the reason: the API can drift from
+  # `src/openapi.json` without anyone opening a pull request.
```
carlsverre: delete comment

.github/workflows/build.yml
```diff
@@ -217,12 +222,63 @@ jobs:
 
       - name: Test
         run: cargo nextest run
+  # The read-only spec suite against a real tenant. The mock cannot catch
+  # `src/openapi.json` drifting from what the API returns, because the mock
+  # fixtures come from the spec. This job also covers real authentication,
+  # cursor pagination, and release-version gating.
```
carlsverre: remove comment

specs/runs.txt
```diff
@@ -230,22 +235,26 @@ snouty --json runs events $R_run_id --match counter
...
+[!staging] mock-runs-server
+[!staging] snouty runs show $R_run_id
+[!staging] stdout 'Failure VTime.*'
+[!staging] stdout 'Failure Hash.*'
+
+[!staging] mock-runs-server
+[!staging] snouty --json runs show $R_run_id
+[!staging] stdout '"failure_moment".*'
+[!staging] stdout '"input_hash".*'
+[!staging] stdout '"vtime".*'
```
carlsverre: move below staging skip

.github/workflows/build.yml
carlsverre: Deleted in `e8cfc25`.

.github/workflows/build.yml
carlsverre: Removed in `e8cfc25`.

specs/runs.txt
carlsverre: Moved in `e8cfc25`. Below the skip the block needs no `[!staging]` prefixes, so those go too. It re-captures `$R_run_id` itself and its order relative to the following blocks is unchanged, so mock execution is identical. Both spec modes verified.

PR #192, opened 2026-08-02
Add a VTime type and show vtime as an exact JSON number

### LINKED ISSUE #191: Add a VTime type and show vtime as an exact JSON number
## Summary

Snouty shows a moment's `vtime` in four different `--json` outputs. Three of them
show a string. One of them shows a number. No type holds the value, and no test
makes sure that the precision is correct.

A `vtime` is precision-sensitive. If you give Antithesis a moment that does not
agree exactly with the value from the API, you get the wrong moment. This is a
bug. The only exception is the `--begin-vtime` field, which is a lower bound.

This issue asks for two changes:

1. Add a `VTime` type. Use it everywhere snouty holds a vtime.
2. Show `vtime` as an exact JSON number in every `--json` output.

A number is better than a string for clients. A client cannot compare strings
alphanumerically by accident, because `"1000.0" < "9.0"` is true but wrong.

## Background: what a vtime is

Antithesis counts virtual time in *ticks*. A tick is one retired CPU tick in the
hypervisor. The API converts ticks to seconds with this formula:

```
vtime_seconds = vtime_ticks / 4294967296     # 4294967296 == 2^32
```

The API always sends the full-precision `f64` representation of that quotient, as
a string. Example: `"398.4898056755774"`.

Snouty used ticks before v0.6.0. The API sent `moment._vtime_ticks`, and the
fault windows used integer ticks. Commit `69d04cf` (2026-06-11) removed this
code. The current spec and the current fixtures do not contain `_vtime_ticks`.

## What we found

We did an investigation before we wrote this issue. Read this section before you
start work. It tells you which problems are real and which problems are not.

### An f64 holds a vtime without a loss of precision

This is true by construction. Three properties together give the guarantee:

1. `ticks / 2^32` has a power-of-two denominator. The division only changes the
   binary exponent. It does not change the mantissa. The result is exact in an
   `f64` for any tick count less than 2^53. That limit is a vtime of 2^21
   seconds, or approximately 24.3 days. Every real run is far below this limit.
2. The `str::parse::<f64>()` function in the Rust standard library is
   correctly-rounded. It recovers the exact `f64` that the API printed.
3. `serde_json` writes an `f64` with ryu, which gives the shortest text that
   parses back to the same `f64`. Go (`'g'` with precision -1) and Python
   (`repr`) print the same text.

### The current conversion is already byte-exact for real data

We tested `str -> f64 -> serde_json` on every full-precision vtime string in the
repository. All 14 of them come back byte-identical, and all of them are
tick-aligned:

| API string | ticks | output |
|---|---|---|
| `398.4898056755774` | 1711500683166 | `398.4898056755774` |
| `313.15126806590706` | 1344974455044 | `313.15126806590706` |
| `22.475549569819123` | 96531750362 | `22.475549569819123` |
| `45.334635781589895` | 194710778058 | `45.334635781589895` |
| `329.8037810830865` | 1416496453849 | `329.8037810830865` |
| `402.0` | 1726576852992 | `402.0` |

We also tested tick recovery at the limits of the range. Tick counts 1, 2^20,
2^32, 2^42, 2^52 and 2^53-1 all give back the exact tick count.

**This is important. There is no large precision bug to repair.** The
`src/moment.rs` path and the `scripts/gen-gallery.py` path both go through an
`f64`, but both are byte-exact for real vtimes. Do not start this work with the
belief that snouty corrupts values today.

An earlier draft of this issue said that the conversion is lossy, because the
fixture at `src/testutils.rs:656` uses `"vtime":"402"` and snouty prints `402.0`.
That fixture is not realistic. The API sends `"402.0"`, which round-trips
correctly.

### A JSON number removes the need for byte-exact text

The requirement for identical text comes only from the string type. After you
change `vtime` to a number, the path is:

```
API f64 -> JSON number -> client f64 -> client text -> API
```

Property 1 makes the `f64` exact. Property 2 makes each parse exact. Property 3
makes each print round-trip. Therefore value fidelity is unconditional. We
confirmed this with a simulated client: it parses the number back to the same
tick count every time.

### The real defects

These are the problems to repair:

1. **The type is not consistent.** `runs logs --json` shows a number
   (`src/runs.rs:1866`). `runs events --json` (`src/runs.rs:1419`), `runs show
   --json` (`src/runs.rs:319`) and `runs properties --json`
   (`src/runs.rs:460`) all show a string.
2. **`src/openapi.json` declares `Moment.vtime` as a `string`.** Log lines refer
   to the `Event` schema. Therefore `runs logs --json` reports the wrong type for
   the API that it proxies.
3. **Nothing enforces the correct conversion.** The rule is in comments only.
   Every new call site can go around it.
4. **No fixture tests precision.** The fault-window fixtures use `"1.0"`,
   `"2.0"` and `"5.0"`. The event fixture uses `"402"`. None of these values can
   fail. This is the reason the type inconsistency stayed hidden.

### Approaches we rejected

**The `arbitrary_precision` feature of `serde_json`.** This feature makes
`Number` hold the original text, and it unlocks `Number::from_string_unchecked`.
We tested it: the build succeeds and 529 of 533 tests pass. (The 4 failures need
`docker` in `PATH` and also fail on a clean tree.) We reject it for these
reasons:

- Cargo features are additive and global. The change to `Number` also reaches
  `jsonschema`, `json-patch`, `progenitor-client` and `http-cache-reqwest`. This
  includes the launch and parameter validation path.
- Number equality becomes textual. `1.0 != 1` and `1.0 != 1.00`.
- A `Value` that goes through a different serializer leaks
  `{"$serde_json::private::Number": "..."}`.
- `from_string_unchecked` needs a strict guard for the JSON number grammar.
  `f64::from_str` accepts `inf`, `NaN` and `1e999`, and none of them are valid
  JSON.
- The three properties above make the feature unnecessary.

**Ticks on the wire.** Ticks give exact integer comparison, and a tick count
below 2^53 is exact even in the doubles that `jq` uses. But the API no longer
sends ticks, and an `f64` is already exact. A tick field adds a second
representation and gives no benefit.

**A string everywhere.** This makes the text exact, but it invites a client to
compare vtimes alphanumerically, which gives the wrong order.

**The `float_roundtrip` feature.** Add this one. It is narrow and safe. The
`as_f64()` fallback in `moment_vtime_seconds` (`src/runs.rs:1606`) lets
`serde_json` parse a vtime, and the default parser of `serde_json` is not
correctly-rounded. This is the only place where snouty can lose one ULP. The
feature changes the parser only. It does not change the representation.

### Step 1: Inject the format marker in `build.rs`

`progenitor` and `typify` support type replacement. We tested this and it works.
`GenerationSettings::with_conversion` maps a schema onto a named type, and the
`example` field in the schema does not stop the match.

Do not edit `src/openapi.json`. That file is a vendored upstream artifact, and
the next refresh of the spec removes the edit.

Instead, add the marker in `build.rs`. That file already loads the spec as a
`serde_json::Value`, and it already asserts invariants on the spec (see
`additional_properties_false_paths`). Add `"format": "vtime"` to
`components.schemas.Moment.properties.vtime`. Assert that the path exists, so an
upstream change to the schema fails the build with a clear message.

Then register the conversion:

```rust
let vtime_schema: schemars::schema::SchemaObject =
    serde_json::from_value(serde_json::json!({"type": "string", "format": "vtime"})).unwrap();
settings.with_conversion(
    vtime_schema,
    "crate::vtime::VTime",
    std::iter::empty::<progenitor::TypeImpl>(),
);
```

Add `schemars = "0.8"` to `[build-dependencies]`. Version 0.8.22 is already in
`Cargo.lock`.

The generated type then becomes:

```rust
pub struct Moment {
    pub input_hash: ::std::string::String,
    pub vtime: crate::vtime::VTime,
}
```

### Step 2: Write `src/vtime.rs`

Give `VTime` these properties:

- It holds one private `f64`.
- `FromStr` uses `str::parse::<f64>()`, because that function is
  correctly-rounded. It rejects a value that is not finite.
- `Deserialize` uses an explicit `Visitor`. Give it `visit_str` and `visit_f64`.
  Reject a value that is not finite in both. Do **not** use
  `#[serde(untagged)]`, because that buffers the value through the `Content`
  type of serde and makes the string path less predictable.
- `Serialize` writes an `f64` with `serialize_f64`.
- `Display` writes the ryu text, so it always agrees with the JSON number. Do
  not use `Number::from_f64().unwrap()`, because that adds a panic path.
- `Eq` and `Ord` use `total_cmp`. This is sound only because `FromStr` and
  `Deserialize` reject a NaN. Write this reason in a comment.
- Add `plus_seconds(f64) -> VTime` for the fault windows. See step 4.
- Add `as_seconds() -> f64` for the human output.
- Add a `ZERO` constant for the placeholder moment check.

Also make `VTime` verify its own premise. Add a method that tests whether
`seconds * 2^32` is an integer. Use it to detect a future API change that breaks
the tick model, instead of a silent absorption of the change.

### Step 3: Repair the typed call sites

The compiler finds all of them. Our experiment gave exactly 6 errors, and all of
them get shorter:

| Site | Now | After |
|---|---|---|
| `src/api.rs:105` | `m.vtime != "0"` | `m.vtime != VTime::ZERO` |
| `src/runs.rs:816` | `sanitize(&event.moment.vtime)` | `Display` |
| `src/runs.rs:823` | `sanitize(&event.moment.vtime)` | `Display` |
| `src/runs.rs:960` | `.parse::<f64>()`, `total_cmp`, plus an order for `None` | `a.moment.vtime.cmp(&b.moment.vtime)` |
| `src/runs.rs:1142` | `moment.vtime.clone()` | `Display` |

`sorted_by_vtime` (`src/runs.rs:957`) becomes approximately 15 lines shorter. The
branches for an unparseable vtime go away, because the type cannot hold one.

### Step 4: Move the fault windows to `VTime`

`FaultWindowBounds` (`src/runs.rs:1904`) holds `start_vtime: f64` and
`end_vtime: Option<f64>`. Change both to `VTime`.

The annotator computes an end for each window, because the API does not send one.
`entry["fault"]["max_duration"]` is the maximum duration from the injector.
`end_vtime = max_duration + latest_vtime` (`src/runs.rs:1810`) is the point where
snouty believes the fault stops. `clear_expired_faults` (`src/runs.rs:1805`) then
drops the window as the stream moves forward. Without an end, one partition shows
as active for the remainder of the run.

Keep this arithmetic exactly as it is. Use `plus_seconds`. Do not change the
computation to `latest - start > duration`. The two forms are not bit-identical
in `f64`, and the test at `src/runs.rs:4380`
(`fault_window_active_at_exact_end_vtime_expires_just_past_end`) pins the
behaviour at the boundary on purpose.

Note that this annotation is already an estimate, because `max_duration` is a
maximum. Snouty over-reports a fault that stopped early. Do not repair this here.

The clock-fault arithmetic does not change. The sum of offsets adds *offsets*,
which are durations, not vtimes. `max_clock_fault_start` needs only `Ord::max`.

### Step 5: Convert the NDJSON paths

`runs logs --json` and `runs events --json` must stay on `serde_json::Value`.

**This is a trap. Read it before you try to use the generated type.** `typify`
generates `Event` as `{ moment: Moment }` only. It does not capture unknown
fields. If you deserialize a log line into `Event`, you lose `output_text`,
`source`, `fault` and every other field. The current code uses `Value` for this
reason.

So give these paths one helper, and use it everywhere:

```rust
fn moment_vtime(entry: &Value) -> Option<VTime>
```

Keep a bare `f64` out of `runs.rs`. Then:

- `runs logs --json`: keep the number. Set it from the `VTime`.
- `runs events --json` (`src/runs.rs:1419`): convert the string to a number.
  **This breaks compatibility.**
- `active_faults` (`src/runs.rs:2035`, `:2043`, `:2063`): keep these numeric.
  `cumulative_offset` stays a number, because it is real arithmetic.

Apply the conversion only on the `--json` path. The human renderers read
`entry["moment"]["vtime"].as_str()` (`src/runs.rs:2197`). If a number reaches
them, they show an empty column.

### Step 6: Keep the human output as it is

`format_log_vtime` (`src/runs.rs:1615`) truncates to 3 decimal places. It cuts
the string directly to keep away from an `f64`. A truncated vtime is for
alignment. It is not a round-trippable moment.

You can now write `truncate_decimals(&vtime.to_string(), 3)`, because `Display`
gives the same text. **Do not use `{:.3}`.** That rounds. Truncation is
deliberate: a vtime that you copy from the screen and paste back as
`--begin-vtime` must land on the line you saw, and never past it. Add a test that
pins this equivalence.

### Step 7: Repair `src/moment.rs`

`Moment.from({ ... vtime: 329.8037810830865 })` comes from the copy-moment button
in the triage report. `src/moment.rs:32` parses it with `json5`. Lines 46-49 then
convert the number to a string with `Number::to_string()`.

This is byte-exact for real vtimes, but it is not enforced. The vtime becomes
`antithesis.debugging.vtime` (`src/api.rs:855`), and snouty sends it as the
moment for a debug session. Route it through `VTime` so the same guarantee holds
here.

`json5` keeps an integer as an `i64` or a `u64` (see `de.rs:544-560`), so
`input_hash` is already safe.

### Step 8: Repair `scripts/gen-gallery.py`

`event_vtime` is a `float` (`:203`). Line 541 sets it with
`float(moment["vtime"])`. Lines 1269, 1282 and 1595 then use it as a positional
moment argument. Hold the string instead. Line 1006 is a `--begin-vtime` lower
bound, so you can leave it as a float.

Type-check the script after the change:

```
uvx pyright scripts/gen-gallery.py
```

### Step 9: Repair the fixtures and the specs

Give the fixtures real full-precision values. Use the table above. Today no
fixture can detect a precision fault.

- `src/testutils.rs:656`: replace `"vtime":"402"`.
- The fault-window fixtures: replace `"1.0"`, `"2.0"` and `"5.0"`.

Update the specs:

- `specs/runs.txt:349-351`: the comment describes the conversion to an `f64` as
  the intended behaviour. Rewrite it to state the guarantee.
- `specs/runs.txt:352-354`: keep the numeric assertions. Add a full-precision
  case.
- `specs/runs.txt:407-408`: `"400.5"` becomes `400.5`, and `"401.75"` becomes
  `401.75`.

Per `AGENTS.md`, write the spec changes first. Three of them break
compatibility.

### Step 10: Update `COOKBOOK.md`

The recipe reads `failure_moment` and gives it to `runs logs`. Add these points:

- `vtime` is a JSON number. Compare it numerically. Never compare it
  alphanumerically.
- Give the value to `runs logs` without a change.
- The check for the placeholder moment is now numeric, not `"0"`.

Verify each command and each `jq` expression by a run, as `AGENTS.md` requires.

## Changes that break compatibility

`vtime` changes from a string to a number in:

- `runs events --json`
- `runs show --json` (`failure_moment.vtime`)
- `runs properties --json`, and the property detail output

`runs logs --json` does not change. It shows a number today.

A client that reads the value with `jq -r` continues to work. A client that
expects a JSON string fails. State this clearly in the pull request and in the
release notes.

Note that this field already changed shape once. Commit `69d04cf` removed the
separate `vtime_seconds` field and changed `moment.vtime` from the ticks of the
server to seconds. That change shipped in v0.6.0.

## Test plan

- A table test that shows each real full-precision vtime comes back
  byte-identical. Use the values in the table above.
- Tick recovery at the limits: 1, 2^20, 2^32, 2^42, 2^52, 2^53-1.
- `FromStr` rejects `NaN`, `inf`, `-inf` and text that is not a number.
- `Deserialize` accepts a string and a number, and rejects a value that is not
  finite in both.
- `Ord` gives a numeric order, and not an alphanumeric order. Test `9.0` against
  `1000.0`.
- The truncation of the human output agrees with the string form. Test that
  truncation cuts and does not round.
- The fault-window boundary tests still pass without a change to their meaning.

Run the full set before you finish:

```
cargo nextest run
cargo clippy
cargo fmt
```

Four tests in `compose::tests` need `docker` in `PATH`. They fail on a clean tree
also. Do not try to repair them.

## Two edge cases

**The placeholder moment.** `"0"` becomes `0.0`. This is the placeholder for an
incomplete run. `real_failure_moment` (`src/api.rs:102-106`) already treats it as
"no moment", so snouty never sends it back. The check in `COOKBOOK.md` becomes
numeric.

**A sub-millisecond vtime.** Below tick 2^20 (approximately 0.000244 seconds),
`serde_json` writes exponent notation, such as `2.3283064365386963e-10`. This is
valid JSON, and it parses back exactly. Go and Python print the same text, so
this is not a regression against the current output. It matters only if a client
gives that text back as a query parameter and the parser of the API rejects `e`
notation. This needs a check against the API.

## Open question

Does the logs endpoint compare the vtime as text, or does it parse the vtime to a
value?

The design in this issue is correct under both answers, because it keeps value
fidelity in every path. So this question does not block the work. But an answer
tells us whether the byte-exact text ever mattered. You can test it against
staging: send a moment as `402.0` where the API sent `402`, and see whether the
logs come back. The same test can show whether the live API still sends
`_vtime_ticks` on a log line, because the schema permits additional properties.

## Files to change

| File | Change |
|---|---|
| `build.rs` | Inject the format marker. Register the conversion. |
| `Cargo.toml` | Add `schemars` to build-dependencies. Add `float_roundtrip`. |
| `src/vtime.rs` | New. The `VTime` type and its tests. |
| `src/lib.rs` | Declare the module. |
| `src/api.rs` | The placeholder check at line 105. |
| `src/runs.rs` | The typed sites, the fault windows, the NDJSON paths, the human output. |
| `src/moment.rs` | Route the `Moment.from` parser through `VTime`. |
| `src/testutils.rs` | Full-precision fixtures. |
| `specs/runs.txt` | The assertions and the comments. |
| `scripts/gen-gallery.py` | Hold the vtime as a string. |
| `COOKBOOK.md` | The numeric type and the numeric placeholder check. |

carlsverre:
A vtime is precision-sensitive: a moment sent back to the API must carry the exact value the API named, or it identifies the wrong moment. Snouty held vtimes as plain strings, and its `--json` surfaces disagreed on the type — `runs logs` printed a number while events, show, and properties printed strings. No type enforced the conversion, and no fixture could detect a precision fault.

This PR adds a `VTime` type and uses it everywhere snouty holds a vtime. The type wraps one private, finite `f64`, which holds any real vtime exactly (a vtime is `ticks / 2^32`, so the division only shifts the binary exponent). Parsing is correctly-rounded and printing is shortest-round-trip, so every path through the type is value-exact — the full test plan from the issue pins this, including byte-identical round trips for every full-precision vtime in the repository and tick recovery up to 2^53−1.

The main changes:

- `build.rs` tags `Moment.vtime` with a `format: vtime` marker and maps it onto `crate::vtime::VTime` with progenitor's `with_conversion`. The vendored `src/openapi.json` is not edited. The build asserts the mapping took, so a spec refresh or generator change cannot silently fall back to a string.
- Every `--json` surface now shows `vtime` as an exact JSON number. The NDJSON stream classifies each line once at its boundary (`NdjsonLine::Entry` — a parsed object with `moment.vtime` already normalized — or `NdjsonLine::Raw`), so no consumer re-parses lines or re-invents the junk-line fallback; commands whose contract is raw passthrough (`runs logs --json --raw`, `runs build-logs --json`) use the raw line streamer and never parse.
- The fault windows (`FaultWindowBounds`) hold `VTime`; the boundary arithmetic and its tests are unchanged. `sorted_by_vtime` shrinks to a `sort_by_key` on the type's `Ord`.
- Human output keeps its shape: the log stream still truncates vtime to 3 decimals (cuts, never rounds — a test pins that the string and number forms agree), and tables show the full-precision `Display` text, which matches the JSON number byte-for-byte.
- `Moment.from` vtimes (number or string) route through `VTime`, so an integer vtime normalizes to the API's own print (`402` → `"402.0"`) and a non-numeric vtime is rejected.
- Fixtures and specs use real full-precision vtimes, chosen so an alphanumeric sort or a one-ULP precision fault fails the suite. `scripts/gen-gallery.py` holds the vtime as the exact number text. `COOKBOOK.md` documents the numeric type and the numeric placeholder check.

**Breaking change:** `vtime` changes from a JSON string to a JSON number in `runs events --json`, `runs show --json` (`failure_moment.vtime`), and `runs properties --json`. A client that reads the value with `jq -r` keeps working; a client that expects a JSON string fails. `runs logs --json` already emitted a number and does not change.

Testing: the mock suite (546 tests), clippy, fmt, and pyright pass, and the staging spec suite passes against orbitinghail. Live probes against orbitinghail confirm the numeric output on real data and answer the issue's open question: the logs endpoint parses the vtime by value, not by text — the exact string, a trailing-zero variant, and `1.297957244166173e1` all stream the identical timeline, so `e` notation is also accepted. A full 83-story gallery generated with the new binary drives every moment-based story through the numeric vtimes end to end.

fixes #191

### DIFF
--- a/COOKBOOK.md
+++ b/COOKBOOK.md
@@ -27,10 +27,13 @@ snouty runs show "$run" --json \
 ```
 
 The moment is `{"input_hash": ..., "vtime": ...}` and feeds `runs logs` in that
-order.
+order. The `vtime` is a JSON number carrying the moment's exact value: pass it
+through unchanged, and compare it numerically — never as text, where
+`"1000.0" < "9.0"`.
 
 Not every run has one. The `failure_moment` key is absent when the run has no
 moment-pinned failure, and a run that timed out or was killed reports the
-placeholder `{"input_hash": "0", "vtime": "0"}` — which streams nothing.
-`snouty runs show` treats that placeholder as "no moment"; do the same, and skip
-the log fetch rather than streaming an empty timeline.
+placeholder `{"input_hash": "0", "vtime": 0}` — which streams nothing.
+`snouty runs show` treats that placeholder as "no moment"; do the same with a
+numeric check (`.vtime != 0`), and skip the log fetch rather than streaming an
+empty timeline.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2741,6 +2741,8 @@ dependencies = [
  "reqwest-middleware",
  "rpassword",
  "rusqlite",
+ "ryu",
+ "schemars",
  "semver",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2741,7 +2741,6 @@ dependencies = [
  "reqwest-middleware",
  "rpassword",
  "rusqlite",
- "ryu",
  "schemars",
  "semver",
  "serde",
@@ -2755,6 +2754,7 @@ dependencies = [
  "toml 1.1.2+spec-1.1.0",
  "which",
  "wiremock",
+ "zmij",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,9 +31,8 @@ serde = { version = "1", features = ["derive"] }
 # float_roundtrip makes serde_json's f64 parsing correctly-rounded, so a vtime
 # read out of a JSON number is never off by one ULP (see src/vtime.rs).
 serde_json = { version = "1", features = ["preserve_order", "float_roundtrip"] }
-# The float printer serde_json itself uses; VTime::Display takes its digits
-# (expanding exponent notation to plain decimal — see src/vtime.rs), so the
-# human text matches the JSON number wherever both are plain decimal.
+# The float printer serde_json itself uses; VTime::Display goes through it so
+# the human text matches the JSON number byte-for-byte (see src/vtime.rs).
 zmij = "1"
 indexmap = { version = "2.10.0" }
 rusqlite = { version = "0.34", features = ["bundled"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,7 @@ serde = { version = "1", features = ["derive"] }
 # float_roundtrip makes serde_json's f64 parsing correctly-rounded, so a vtime
 # read out of a JSON number is never off by one ULP (see src/vtime.rs).
 serde_json = { version = "1", features = ["preserve_order", "float_roundtrip"] }
-# The shortest-round-trip float printer serde_json itself uses; VTime's
-# Display goes through it so human output matches the JSON number exactly.
+# Shortest-round-trip float printing for VTime::Display (see src/vtime.rs).
 ryu = "1"
 indexmap = { version = "2.10.0" }
 rusqlite = { version = "0.34", features = ["bundled"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,9 @@ serde = { version = "1", features = ["derive"] }
 # float_roundtrip makes serde_json's f64 parsing correctly-rounded, so a vtime
 # read out of a JSON number is never off by one ULP (see src/vtime.rs).
 serde_json = { version = "1", features = ["preserve_order", "float_roundtrip"] }
-# The float printer serde_json itself uses; VTime::Display goes through it so
-# the human text matches the JSON number byte-for-byte (see src/vtime.rs).
+# The float printer serde_json itself uses; VTime::Display takes its digits
+# (expanding exponent notation to plain decimal — see src/vtime.rs), so the
+# human text matches the JSON number wherever both are plain decimal.
 zmij = "1"
 indexmap = { version = "2.10.0" }
 rusqlite = { version = "0.34", features = ["bundled"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,12 @@ color-eyre = { version = "0.6", default-features = false }
 futures-util = "0.3"
 tokio = { version = "1", features = ["rt", "macros", "process", "time", "io-util", "io-std", "signal", "sync"] }
 serde = { version = "1", features = ["derive"] }
-serde_json = { version = "1", features = ["preserve_order"] }
+# float_roundtrip makes serde_json's f64 parsing correctly-rounded, so a vtime
+# read out of a JSON number is never off by one ULP (see src/vtime.rs).
+serde_json = { version = "1", features = ["preserve_order", "float_roundtrip"] }
+# The shortest-round-trip float printer serde_json itself uses; VTime's
+# Display goes through it so human output matches the JSON number exactly.
+ryu = "1"
 indexmap = { version = "2.10.0" }
 rusqlite = { version = "0.34", features = ["bundled"] }
 console = "0.15"
@@ -60,6 +65,9 @@ openapiv3 = "2"
 prettyplease = "0.2"
 progenitor = "0.14.0"
 quote = "1"
+# Same schemars version progenitor/typify resolve; build.rs hands
+# with_conversion a schemars SchemaObject to map vtime onto crate::vtime::VTime.
+schemars = "0.8"
 serde_json = "1"
 syn = "2"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,9 @@ serde = { version = "1", features = ["derive"] }
 # float_roundtrip makes serde_json's f64 parsing correctly-rounded, so a vtime
 # read out of a JSON number is never off by one ULP (see src/vtime.rs).
 serde_json = { version = "1", features = ["preserve_order", "float_roundtrip"] }
-# Shortest-round-trip float printing for VTime::Display (see src/vtime.rs).
-ryu = "1"
+# The float printer serde_json itself uses; VTime::Display goes through it so
+# the human text matches the JSON number byte-for-byte (see src/vtime.rs).
+zmij = "1"
 indexmap = { version = "2.10.0" }
 rusqlite = { version = "0.34", features = ["bundled"] }
 console = "0.15"
--- a/build.rs
+++ b/build.rs
@@ -143,9 +143,7 @@ fn untype_error_responses(spec: &mut serde_json::Value) {
 /// `with_conversion` mapping registered above. The marker is injected here
 /// rather than edited into `src/openapi.json`, because that file is a
 /// vendored upstream artifact — the next spec refresh would silently drop the
-/// edit. Panics when the spec no longer carries the expected string-typed
-/// property, so an upstream schema change fails the build with a clear
-/// message instead of silently regenerating a stringly-typed client.
+/// edit.
 fn mark_vtime_schema(spec: &mut serde_json::Value) {
     let vtime = spec
         .pointer_mut("/components/schemas/Moment/properties/vtime")
--- a/build.rs
+++ b/build.rs
@@ -42,17 +42,39 @@ fn generate_api_client(out_dir: &Path) {
         offenders.join(", ")
     );
     untype_error_responses(&mut spec_value);
+    mark_vtime_schema(&mut spec_value);
     let spec: openapiv3::OpenAPI = serde_json::from_value(spec_value).unwrap();
 
     let mut settings = progenitor::GenerationSettings::default();
     settings.with_interface(progenitor::InterfaceStyle::Builder);
     settings.with_inner_type(quote::quote!(crate::api::ClientState));
+    // Map the marked vtime schema onto the handwritten VTime type, which
+    // enforces the exact string<->f64 conversion a vtime needs (the
+    // conversion lookup ignores schema metadata such as description/example,
+    // so `type` + `format` is the whole match key).
+    let vtime_schema: schemars::schema::SchemaObject =
+        serde_json::from_value(serde_json::json!({"type": "string", "format": "vtime"})).unwrap();
+    settings.with_conversion(
+        vtime_schema,
+        "crate::vtime::VTime",
+        std::iter::empty::<progenitor::TypeImpl>(),
+    );
     let mut generator = progenitor::Generator::new(&settings);
     let tokens = generator.generate_tokens(&spec).unwrap();
     let ast = syn::parse2(tokens).unwrap();
     let content = prettyplease::unparse(&ast);
     let content = patch_lenient_booleans(content);
     assert_no_typed_error_responses(&content);
+
+    // The conversion fails open: an unmatched schema silently falls back to a
+    // plain String field. Assert it took, so a progenitor/typify change that
+    // breaks the match fails the build instead of quietly shipping a client
+    // without the vtime precision guarantees.
+    assert!(
+        content.contains("pub vtime: crate::vtime::VTime"),
+        "generated client does not use crate::vtime::VTime for Moment.vtime; \
+         the with_conversion schema match no longer applies"
+    );
 
     fs::write(out_dir.join("antithesis_api.rs"), content).unwrap();
 }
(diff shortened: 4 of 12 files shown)


carlsverre:
@claude re-review changes

carlsverre:
@claude review

jeskew (approved):
LGTM

PR #193, opened 2026-08-02
Polish failure output: render concerns at print time

carlsverre:
Stacked on #183 — the base is `status-first-api-errors`, and only the top commit (`7ead804`) is this PR. Retarget to `main` after #183 merges.

A satisfaction review of the failure-behaviour gallery stories (21 stories: API errors, empty results, validate failures, login error/repair paths) found five issues. This PR fixes them, with one rule throughout: wrapping and frame layout are properties of *printing*, so they live at the print sites and messages stay single logical lines.

## 1. A single error no longer wears a `0:` frame

color_eyre numbers every cause chain, so the common case — one error — rendered as:

```
Error:
   0: API error: 404 Not Found — Resource not found
```

The `0:` enumerates nothing and reads like a stack frame. `error::render_report` is now the one place a `Report` becomes terminal text (called from `main`); it collapses a one-element chain onto the `Error:` line:

```
Error: API error: 404 Not Found — Resource not found
```

A chain of two or more keeps its numbering, which is genuinely counting causes there:

```
Error:
   0: failed to contact API
   1: error sending request for url (…)
```

`specs/runs.txt` pins the collapsed form (`^Error: API error…`, and no `0:` frame).

## 2. Overlong prose wraps at print time — when a person is reading

`render::wrap_text` wraps at 100 columns: word-boundary, ANSI-aware (escape sequences count as zero width), continuation lines keep their indent, and words are never split — a path overflows rather than breaking. **A line that already fits is returned byte-identical**, so aligned content — doctor's settings table, the TOML caret marker in login's repair prompt — is never re-flowed.

Wrapping applies **only when stderr is a terminal** (`wrap_if(text, stderr().is_terminal())`), the same rule the color theme already follows at the same print sites. The first draft wrapped unconditionally, and CI caught why that is wrong: the wrap point moves with the length of an embedded path, so any multi-word assertion that straddles it breaks — `launch_config_rejects_ambiguous_dir` failed on macOS and the nix runner (long temp dirs) while passing on Linux (short `/tmp`). Reproduced by sweeping `TMPDIR` lengths: 56–76-column paths split the asserted phrase. Pipes and captures now get one stable logical line at any path length; a pty run shows the tenant error wrapping at 100 columns with colors intact.

Sites: `main`'s report printing, the stranded-scripts warning (was 215 columns on a terminal), and login's failed-to-load and proceed prompts (was 230).

The two corrupt-file recovery notes (settings and credentials) are restructured instead of wrapped — each path moves to its own line, because a path is one unsplittable word and the paths are what the user copies:

```
note: the existing settings file could not be parsed; a new one will be written.
  kept as a backup: /home/you/.config/snouty/settings.toml.bak
  will be rewritten: /home/you/.config/snouty/settings.toml
```

These print on a *recovery* path (login continues and exits 0), so there is no `Report` in flight for `.note()` sections to attach to; the multi-line `eprintln!` is the hand-rolled equivalent of that rendering.

## 3. `runs logs` on a bad moment names the next command

Measured against the live tenant: a bad run id already prints `run not found: <id>`, because `explain_run_scoped_error` probes `get_run` on every run-scoped 404. So a surviving 404 always means *the run exists and the moment does not* — and the message now says so:

```
Error: API error: 404 Not Found — Resource not found

Suggestion: the run exists but no moment matches this hash and vtime — list valid moments with
`snouty runs events <run-id>`
```

No added request: the new `explain_logs_error` reuses the probe that already ran. The mock's logs route is not moment-aware, so this is covered by wiremock unit tests rather than a spec.

## 4. Three validate errors now name their fix

Matching the standard their siblings set (`wrong-extension` suggests the exact rename, `missing-image` says "pull or build"):

- unrecognized command names → `Suggestion: rename each command to start with a recognized prefix (first_, driver_, anytime_, eventually_, finally_) or helper_`
- non-executable commands → `Suggestion: chmod +x the listed commands in the image, then rebuild`
- external network → `Suggestion: remove `external: true` and declare the network normally — Antithesis provisions every network inside the test environment`

Pinned in `specs_engine/validate_failures.txt` and a compose unit test.

## 5. Gallery login transcripts preserve interaction order

`gen-gallery.py` captured stdout and stderr separately and concatenated them, so the repair story showed "Saved tenant…" *before* the proceed-prompt that actually preceded it. Login stories now capture with stderr merged into stdout. pyright: 0 errors.

## Two CI changes this PR also carries

**The build workflow runs on every pull request.** The `branches: ["main"]` filter meant this stacked PR (base = #183's branch) got no gate at all on its first run. Dropping the filter gave it — and every future stacked PR — the full matrix, which is what then caught the wrap bug above.

**The staging event searches are bounded.** The full gate exposed that `runs events $R_run_id --match request` hung CI to its 15-minute timeout: the suite tests the tenant's newest completed run, the needle is common, and an unbounded search streams every match — wall time scaled with whatever run the tenant produced last (16s two days ago, 70s locally, >12 minutes in CI). The event-search assertions are all `[!staging]` mock pins, so staging now runs `--limit 3` twins and the mock keeps the unbounded form. Staging suite: 94s → 22s; the hot search: 70s → 2.5s.

## Measured effect

Regenerated gallery stories on the unconditional-wrap draft showed the terminal experience (login-repair-broken 244→134 columns, login-bad-tenant 230→100, validate-stranded 215→99). With tty-conditional wrapping, gallery captures (piped) show the unwrapped single lines again — the wrap benefit is real but only visible on a terminal. What captures do keep: the collapsed `Error:` frame, the moment suggestion on `runs-logs-bad-moment`, the two-path backup notes, and the three validate suggestions.

## Testing

`cargo test` (445 lib tests plus every integration and spec target), `clippy --all-targets`, `fmt --check`, and `typos` are clean. Spec suite green in both modes, including staging against a live tenant, and the `cli_launch_config` suite passes at every `TMPDIR` length in the sweep. New unit tests: `wrap_text` (4), `render_report` (4), `explain_logs_error` (2), plus the extended compose external-network test.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -54,8 +54,8 @@ impl std::error::Error for ApiError {}
 ///
 /// This is the one place that turns a `Report` into terminal text, so printing
 /// concerns live here and not in the messages: the chain index is dropped when
-/// there is nothing to enumerate, and overlong prose wraps to
-/// [`crate::render::PROSE_WIDTH`].
+/// there is nothing to enumerate. Wrapping is the caller's call — it depends
+/// on whether stderr is a terminal, which the print site knows.
 ///
 /// color_eyre renders every report as a numbered chain, so a single error —
 /// the common case — arrives as `\n   0: <message>`. The `0:` numbers
@@ -69,7 +69,7 @@ pub fn render_report(report: &Report) -> String {
     } else {
         rendered
     };
-    crate::render::wrap_text(&format!("Error: {rendered}"), crate::render::PROSE_WIDTH)
+    format!("Error: {rendered}")
 }
 
 /// Remove the `\n   0: ` frame header from a rendered single-element chain,
@@ -160,14 +160,15 @@ mod tests {
         assert!(rendered.contains("1: "), "got: {rendered}");
     }
 
+    // Wrapping belongs to the print site (it is tty-dependent), so the
+    // renderer must hand back the message as one logical line.
     #[test]
-    fn render_report_wraps_overlong_prose() {
-        let report = user_error(format!("prefix {}", "word ".repeat(40)));
+    fn render_report_does_not_wrap() {
+        let long = format!("prefix {}", "word ".repeat(40));
+        let report = user_error(long.clone());
         let rendered = plain(&render_report(&report));
-        let message_lines: Vec<&str> = rendered.lines().take_while(|l| !l.is_empty()).collect();
-        assert!(message_lines.len() > 1, "got: {rendered}");
         assert!(
-            message_lines.iter().all(|l| l.len() <= 100),
+            rendered.lines().next().unwrap().contains(&long),
             "got: {rendered}"
         );
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -53,49 +53,26 @@ impl std::error::Error for ApiError {}
 /// Render a report for the user, at the moment of printing.
 ///
 /// This is the one place that turns a `Report` into terminal text, so printing
-/// concerns live here and not in the messages: the chain index is dropped when
-/// there is nothing to enumerate. Wrapping is the caller's call — it depends
-/// on whether stderr is a terminal, which the print site knows.
+/// concerns live here and not in the messages. Wrapping is the caller's call —
+/// it depends on whether stderr is a terminal, which the print site knows.
 ///
 /// color_eyre renders every report as a numbered chain, so a single error —
-/// the common case — arrives as `\n   0: <message>`. The `0:` numbers
-/// nothing and reads like a stack frame; collapse it onto the `Error:` line.
-/// A report with two or more chain elements keeps its numbering, which is
-/// genuinely enumerating causes there.
+/// the common case — arrives as `\n   0: <message>`. The `0:` numbers nothing
+/// and reads like a stack frame. For a one-element chain the message comes
+/// from the report itself instead, onto the `Error:` line. The sections a
+/// report carries (`Suggestion:`, `Note:`, payload snippets) render only
+/// through the handler, as the text after the first blank line; that tail is
+/// kept verbatim. A chain of two or more keeps color_eyre's numbering, which
+/// is genuinely enumerating causes there.
 pub fn render_report(report: &Report) -> String {
     let rendered = format!("{report:?}");
-    let rendered = if report.chain().len() == 1 {
-        collapse_single_frame(&rendered)
-    } else {
-        rendered
-    };
-    format!("Error: {rendered}")
-}
-
-/// Remove the `\n   0: ` frame header from a rendered single-element chain,
-/// tolerating ANSI color codes around the index. Continuation lines keep the
-/// six-column indent color_eyre gave them.
-fn collapse_single_frame(rendered: &str) -> String {
-    let stripped = rendered.trim_start_matches('\n');
-    // Optional escape sequence, the literal `0`, optional escape, `:`.
-    fn skip_ansi(mut value: &str) -> &str {
-        while let Some(after) = value.strip_prefix('\u{1b}') {
-            match after.find(|c: char| c.is_ascii_alphabetic()) {
-                Some(i) => value = &after[i + 1..],
-                None => break,
-            }
-        }
-        value
+    if report.chain().len() > 1 {
+        return format!("Error: {rendered}");
     }
-
-    let rest = skip_ansi(stripped.trim_start_matches(' '));
-    if let Some(after) = rest.strip_prefix('0')
-        && let Some(after) = skip_ansi(after).strip_prefix(':')
-    {
-        return after.trim_start_matches(' ').to_string();
+    match rendered.split_once("\n\n") {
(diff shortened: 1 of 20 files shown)


src/error.rs
```diff
@@ -50,6 +50,54 @@ impl std::fmt::Display for ApiError {
...
+    } else {
+        rendered
+    };
+    format!("Error: {rendered}")
+}
+
+/// Remove the `\n   0: ` frame header from a rendered single-element chain,
+/// tolerating ANSI color codes around the index. Continuation lines keep the
+/// six-column indent color_eyre gave them.
+fn collapse_single_frame(rendered: &str) -> String {
```
carlsverre: rather than parsing the rendered output, why not just render directly from the Report.chain if len == 1?

src/render.rs
```diff
@@ -55,6 +55,94 @@ enum NewlinePolicy {
...
+pub(crate) const PROSE_WIDTH: usize = 100;
+
+/// Wrap `text` for stderr when `tty` says a person is reading it.
+///
+/// Piped and captured output stays unwrapped: tests, specs, and shell
+/// pipelines consume whole lines, and a wrap point that moves with an
+/// embedded path length breaks any multi-word match that straddles it. The
+/// caller passes its own terminal check, so the tty decision stays at the
+/// print site.
+pub fn wrap_if(text: &str, tty: bool) -> String {
```
carlsverre: kill this function in favor of a wrap_if_tty function that internalizes the tty check - I'm assuming you wrote it this way for unit tests, but you don't need to unit test this function. it's trivial

src/render.rs
```diff
@@ -55,6 +55,94 @@ enum NewlinePolicy {
...
+/// report wraps the same as a plain one. Words longer than the width (paths,
+/// URLs) are left intact and overflow rather than being split.
+pub(crate) fn wrap_text(text: &str, width: usize) -> String {
+    text.lines()
+        .map(|line| wrap_line(line, width))
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+fn wrap_line(line: &str, width: usize) -> String {
```
carlsverre: honestly this is the kind of thing I'd prefer to use a dependency for if there is one that's reasonable

src/validate.rs
```diff
@@ -696,6 +703,16 @@ fn combined_discovery_error(
         });
     }
 
+    if !unrecognized.is_empty() {
+        err = err.suggestion(
+            "rename each command to start with a recognized prefix (first_, driver_, anytime_, \
```
carlsverre: the prefix list here should match what scan does. and the other claude review is correct - driver_ is not a valid prefix. singleton_driver, parallel_driver, and serial_driver are

carlsverre: Fixed in `dd402c5`, at the depth you asked for: the prefixes now live in one table in scripts.rs (`RECOGNIZED_PREFIXES`) that drives both `ScriptType::from_prefix` and the suggestion text, so the two cannot drift. The rendered suggestion now reads:

```
Suggestion: rename each command to start with a recognized prefix (first_, parallel_driver_, serial_driver_, singleton_driver_, anytime_, eventually_, finally_) or helper_
```

src/render.rs
carlsverre: Done in `dd402c5` — `textwrap` 0.16, default features off. It brings exactly the two pieces I hand-rolled: `core::display_width` skips ANSI escape sequences, and `break_words(false)` + `WordSplitter::NoHyphenation` keep paths whole. The one piece that stays ours is the fits-stays-byte-identical guard, so aligned content (doctor table, TOML caret) is never re-flowed — textwrap would collapse internal spacing on lines it touches. `visible_width` and the greedy wrapper are deleted.

src/render.rs
carlsverre: Done in `dd402c5`. `wrap_if_tty(text)` internalizes the `stderr().is_terminal()` check; callers stop threading a boolean, and the gate has no test. You guessed the motive right — and it was doubly moot, because under `cargo test` stderr is piped, so the wrap path was never reachable through the gate anyway. The two remaining tests sit on the inner pure function and pin our policy choices (byte-identical short lines, hanging indent, no word splitting).

src/error.rs
carlsverre: Done in `dd402c5`, with one constraint that shaped it. The message itself now comes straight from the report (`Display`) when the chain has one element — no frame parsing, and the ANSI-tolerant parser is deleted. What cannot come from the chain is the sections: `Suggestion:`, `Note:`, and payload snippets live in color_eyre's handler and render only through `Debug`, with no public accessor. So the rendering keeps the `Debug` tail after the first blank line verbatim and replaces only the head. One visible trade: the message loses color_eyre's red on a terminal, because `Display` is plain.

carlsverre: Red restored in the follow-up commit: on a terminal the message paints bright red via `color_eyre::owo_colors` (the same `[91m` the handler used), and piped output stays byte-exact plain. The check sits next to the wrap check, for the same reason.

PR #195, opened 2026-08-03
Add an unstable update channel

### LINKED ISSUE #189: snouty beta channel for updates
I want to be able to ship snouty more aggressively to folks ok with sometimes seeing a breaking version of snouty. The idea is to add a beta setting that causes the update command to look for a more recent pre-release version of snouty and use that instead of the latest release. The command should take a channel = beta/release flag to override the decision if someone needs to downgrade to the latest release due to a bug (or update to beta without setting the setting). The setting should still use latest if latest is the latest version available (i.e., if it finds a pre-release that is older than latest, it should use latest)

shipping updates will require possibly tweaking how we ship snouty to support shipping to a github pre-release rather than full release. we will need to tag the versions with `.rcXX` which will hopefully ensure that we don't accidentally send out a full release.

carlsverre:
This adds an unstable update channel so we can ship pre-releases to users who opt in.

A new `update_channel` setting (`SNOUTY_UPDATE_CHANNEL`, or `update_channel` in a settings file) selects the channel: `stable` (the default) or `unstable`. The setting resolves to a typed `UpdateChannel` during settings resolution, so an invalid value fails at startup like any other malformed setting. `UpdateChannel` owns the string conversion via `FromStr` and named consts. `snouty update` gains a `--channel stable|unstable` flag that overrides the setting for one run.

On the unstable channel, snouty passes `--prerelease` to the bundled `snouty-update` helper. The helper (axoupdater) picks the greatest version across releases and pre-releases, and semver orders `0.7.0-rc.1` before `0.7.0` — so when the latest release is newer than every pre-release, the unstable channel still installs the release. An explicit version argument bypasses the channel, as before.

Shipping needs no CI change: the cargo-dist workflow already marks a tag with a pre-release suffix as a GitHub pre-release. The release skill accepts `-rc.N` versions and picks the next rc number from existing tags.

Other changes:

- `snouty doctor` reports the resolved update channel in the settings table.
- `snouty update` now resolves settings, so a broken settings file fails it up front like other commands (previously it ignored settings entirely).
- `Settings::for_test` is now a test-only builder, so test call sites name only the values they set instead of growing a positional argument per setting.
- Specs cover the flag, the setting, and validation of both; unit tests cover channel parsing, settings plumbing, and the doctor row.

fixes #189

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -1,19 +1,26 @@
 ---
 name: Release Snouty
-description: This skill should be used when the user asks to "release snouty", "cut a release", "bump the version", "create a release", or provides a version like "release snouty v0.2.0". Handles version validation, Cargo.toml bump, build, test, commit, and tagging.
+description: This skill should be used when the user asks to "release snouty", "cut a release", "cut a pre-release", "cut an rc", "bump the version", "create a release", or provides a version like "release snouty v0.2.0" or "release snouty v0.7.0-rc.1". Handles version validation, Cargo.toml bump, build, test, commit, and tagging, for releases and pre-releases.
 ---
 
 # Release Snouty
 
 Perform a versioned release of snouty by bumping `Cargo.toml`, building, testing, committing, and tagging. _Do not_ push the resulting commit so the user has a chance to audit it first.
 
+## Pre-releases
+
+A version with an `-rc.N` suffix (e.g. `0.7.0-rc.1`) is a pre-release. The procedure is the same as for a release; only these points differ:
+
+- When the user asks for a pre-release without a full version (e.g. "cut an rc for 0.7.0"), pick the next rc number: list existing tags with `git tag -l 'v0.7.0-rc.*'` and use N+1 of the highest, or `-rc.1` when there are none.
+- Shipping is automatic. Pushing the tag triggers the cargo-dist workflow (`.github/workflows/release.yml`), which builds the artifacts and creates the GitHub release. A pre-release suffix in the tag makes cargo-dist mark it as a GitHub **pre-release**, so `snouty update` ignores it unless the user is on the beta channel (`update_channel = "beta"` or `snouty update --channel beta`). Never create the GitHub release by hand — a hand-made release could miss the pre-release flag and ship the rc to everyone.
+- Do not publish a pre-release to crates.io (see step 7).
+- A later full release of the same version (e.g. `0.7.0` after `0.7.0-rc.2`) is a normal upgrade: semver sorts every `-rc.N` before the release.
+
 ## Release Procedure
 
 ### 1. Parse and Validate the Version
 
-Extract the version from the user's input. Accept formats like `v0.2.0` or `0.2.0`. Strip the leading `v` to get the bare semver.
-
-A version with an `-rc.N` suffix (e.g. `0.7.0-rc.1`) is a pre-release. The release workflow marks the GitHub release as a pre-release automatically, so `snouty update` only installs it for users on the beta channel.
+Extract the version from the user's input. Accept formats like `v0.2.0` or `0.2.0`, including pre-release forms like `v0.7.0-rc.1`. Strip the leading `v` to get the bare semver.
 
 Run all of the following sanity checks before making any changes:
 
@@ -69,4 +76,4 @@ git push && git push --tags
 cargo publish
 ```
 
-For a pre-release (`-rc.N`), omit `cargo publish` — pre-releases ship only as GitHub pre-releases for the beta update channel.
+For a pre-release (`-rc.N`), omit `cargo publish` — pre-releases ship only as GitHub pre-releases for the beta update channel. Also tell the user that after the tag push, CI creates the GitHub pre-release automatically and only beta-channel users receive it.
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -9,25 +9,20 @@ Perform a versioned release of snouty by bumping `Cargo.toml`, building, testing
 
 ## Pre-releases
 
-A version with an `-rc.N` suffix (e.g. `0.7.0-rc.1`) is a pre-release. The procedure is the same as for a release; only these points differ:
-
-- When the user asks for a pre-release without a full version (e.g. "cut an rc for 0.7.0"), pick the next rc number: list existing tags with `git tag -l 'v0.7.0-rc.*'` and use N+1 of the highest, or `-rc.1` when there are none.
-- Shipping is automatic. Pushing the tag triggers the cargo-dist workflow (`.github/workflows/release.yml`), which builds the artifacts and creates the GitHub release. A pre-release suffix in the tag makes cargo-dist mark it as a GitHub **pre-release**, so `snouty update` ignores it unless the user is on the beta channel (`update_channel = "beta"` or `snouty update --channel beta`). Never create the GitHub release by hand — a hand-made release could miss the pre-release flag and ship the rc to everyone.
-- `cargo publish` works for pre-releases too, and is safe: cargo only selects a pre-release when a user asks for it explicitly (e.g. `cargo install snouty --version 0.7.0-rc.1`), so stable users never receive it.
-- A later full release of the same version (e.g. `0.7.0` after `0.7.0-rc.2`) is a normal upgrade: semver sorts every `-rc.N` before the release.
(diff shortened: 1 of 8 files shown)


.claude/skills/release/SKILL.md
```diff
@@ -1,23 +1,32 @@
...
+
+- When the user asks for a pre-release without a full version (e.g. "cut an rc for 0.7.0"), pick the next rc number: list existing tags with `git tag -l 'v0.7.0-rc.*'` and use N+1 of the highest, or `-rc.1` when there are none.
+- Shipping is automatic. Pushing the tag triggers the cargo-dist workflow (`.github/workflows/release.yml`), which builds the artifacts and creates the GitHub release. A pre-release suffix in the tag makes cargo-dist mark it as a GitHub **pre-release**, so `snouty update` ignores it unless the user is on the beta channel (`update_channel = "beta"` or `snouty update --channel beta`). Never create the GitHub release by hand — a hand-made release could miss the pre-release flag and ship the rc to everyone.
+- `cargo publish` works for pre-releases too, and is safe: cargo only selects a pre-release when a user asks for it explicitly (e.g. `cargo install snouty --version 0.7.0-rc.1`), so stable users never receive it.
+- A later full release of the same version (e.g. `0.7.0` after `0.7.0-rc.2`) is a normal upgrade: semver sorts every `-rc.N` before the release.
```
carlsverre: these followon notes are not useful to the skill. focus on actionable choices and actions the skill has to make/do, not background on what else happens downstream.

.claude/skills/release/SKILL.md
```diff
@@ -1,23 +1,32 @@
...
+
 ## Release Procedure
 
 ### 1. Parse and Validate the Version
 
-Extract the version from the user's input. Accept formats like `v0.2.0` or `0.2.0`. Strip the leading `v` to get the bare semver.
+Extract the version from the user's input. Accept formats like `v0.2.0` or `0.2.0`, including pre-release forms like `v0.7.0-rc.1`. Strip the leading `v` to get the bare semver.
```
carlsverre:
```suggestion
Extract the version from the user's input. Accept formats like `v0.2.0`, `0.2.0`, or `v0.7.0-rc.1`. Strip the leading `v` to get the bare semver. If `-rc.N` is specified, this is a pre-release.
```

.claude/skills/release/SKILL.md
```diff
@@ -1,23 +1,32 @@
...
 
 Run all of the following sanity checks before making any changes:
 
-- Validate the version matches `MAJOR.MINOR.PATCH` where each component is a non-negative integer.
+- Validate the version matches `MAJOR.MINOR.PATCH` where each component is a non-negative integer, optionally followed by `-rc.N` where N is a positive integer.
 - Read `Cargo.toml` and extract the current version.
-- Confirm the new version is strictly greater than the current version (compare major, then minor, then patch).
+- Confirm the new version is strictly greater than the current version under semver ordering (compare major, then minor, then patch; a pre-release sorts before its release, so `0.7.0-rc.1` < `0.7.0`, and `0.7.0-rc.2` > `0.7.0-rc.1`).
 - Confirm the git tag `vX.Y.Z` does not already exist (`git tag -l vX.Y.Z`).
```
carlsverre:
```suggestion
- Confirm the git tag `vX.Y.Z[-rc.N]` does not already exist (`git tag -l vX.Y.Z[-rc.N]`).
```

carlsverre: is this clear?

.claude/skills/release/SKILL.md
```diff
@@ -66,3 +75,5 @@ Tell the user to run the following once satisfied:
 git push && git push --tags
 cargo publish
 ```
+
+For a pre-release (`-rc.N`), also tell the user that after the tag push, CI creates the GitHub pre-release automatically and only beta-channel users receive it via `snouty update`. Publishing the pre-release crate is safe: cargo never selects a pre-release unless a user requests it explicitly.
```
carlsverre: delete this

src/cli.rs
```diff
@@ -362,6 +369,19 @@ pub struct UpdateArgs {
...
+    pub channel: Option<UpdateChannel>,
+}
+
+/// Which releases `snouty update` considers when no explicit version is given.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, ValueEnum)]
+pub enum UpdateChannel {
+    /// Install the latest release
+    Release,
+    /// Also consider pre-releases, but prefer the latest release when it is newer
+    Beta,
```
carlsverre: lets rename the channels to stable/unstable

carlsverre: make this consistent throughout the PR + the PR description/title

src/doctor.rs
```diff
@@ -593,41 +598,69 @@ mod tests {
...
+            None, None, None, None, None, None, None,
+        ));
         assert_eq!(row(&rows, "tenant").value, "not set");
     }
 
     #[test]
     fn https_proxy_row_shows_value() {
-        let settings =
-            Settings::for_test(None, None, None, None, Some("http://proxy.corp:8080"), None);
+        let settings = Settings::for_test(
```
carlsverre: refactor Settings::for_test into a test-only builder pattern to stop all these functions calls getting bigger every time we add a setting (and make the tests more readable)

src/main.rs
```diff
@@ -408,19 +408,46 @@ fn check_update_target(requested: &str, current: &str, force: bool) -> Result<()
...
+fn resolve_update_channel(
+    flag: Option<UpdateChannel>,
+    setting: Option<&str>,
+) -> Result<UpdateChannel> {
+    if let Some(channel) = flag {
+        return Ok(channel);
+    }
+    match setting {
+        None | Some("release") => Ok(UpdateChannel::Release),
+        Some("beta") => Ok(UpdateChannel::Beta),
```
carlsverre: the only place we should be converting strings to UpdateChannels is *in* UpdateChannel. add a FromStr impl to UpdateChannel. UpdateChannel should also define const or static strs which can be used in place of string literals for defaults or example lists.

src/main.rs
```diff
@@ -408,19 +408,46 @@ fn check_update_target(requested: &str, current: &str, force: bool) -> Result<()
...
+    // (pre-releases included), so --prerelease is only needed when the beta
+    // channel picks "latest". The helper then installs the greatest version
+    // across releases and pre-releases, so a release newer than every
+    // pre-release still wins.
     let mut updater = Command::new("snouty-update");
     if let Some(version) = &args.version {
         updater.arg("--version").arg(version);
+    } else if channel == UpdateChannel::Beta {
+        updater.arg("--prerelease");
     }
```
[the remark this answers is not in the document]
carlsverre: I disagree with claude here - I think it's fine to push validation early.

src/settings.rs
```diff
@@ -82,6 +83,7 @@ pub struct Settings {
     base_url: Option<String>,
     https_proxy: Option<String>,
     container_engine: Option<String>,
+    update_channel: Option<String>,
```
carlsverre: This should be `update_channel: UpdateChannel` -> fully resolve it during setting resolution as the baseline default.

.claude/skills/release/SKILL.md
carlsverre: Trimmed the pre-release section to the one actionable step (picking the next rc number) and dropped the downstream notes.

.claude/skills/release/SKILL.md
carlsverre: Applied.

.claude/skills/release/SKILL.md
carlsverre: Applied — clear.

.claude/skills/release/SKILL.md
carlsverre: Deleted.

src/cli.rs
carlsverre: Renamed to stable/unstable throughout the code, specs, docs, skill, and the PR title/description.

src/doctor.rs
carlsverre: Done — `Settings::for_test()` now returns a test-only `SettingsBuilder`; call sites name only the values they set.

src/main.rs
carlsverre: Done — `UpdateChannel` owns parsing via `FromStr` and defines `STABLE`/`UNSTABLE` consts used for defaults and error messages.

src/settings.rs
carlsverre: Done — `Settings.update_channel` is a resolved `UpdateChannel` defaulting to `Stable`; an invalid value now fails during settings resolution.

src/settings.rs
carlsverre, line 84: derive Default and use it rather than settings builder -> build for when you just want to test default settings

carlsverre: Done — `Settings` derives `Default` and tests use `Settings::default()` where nothing is set. Also per chat: the builder lost its `cfg(test)` and is now regular public API behind `Settings::builder()`.

PR #200, opened 2026-08-04
Add support for filter_logs_matching in snouty CLI

agupps:
Add support to pass `--filter-logs-matching` to snouty launch.

The filter_logs_matching parameter is a regular expression evaluated by RE2 that is applied on every log message that is outputted during the test run. If the log matches the regular expression, the log is "thrown out" and not stored in the database and does not influence fuzzing in the run. However, the logs are recoverable if the moment is re-warmed in multi-verse debugging sessions.

### DIFF
--- a/specs/launch.txt
+++ b/specs/launch.txt
@@ -137,3 +137,9 @@ stderr 'API error: 404 Not Found.*'
 mock-server 200 '{"runId":"run-123","statusCode":202}'
 snouty run -w basic_test --duration 30
 stderr 'snouty run.*is deprecated.*snouty launch'
+
+mock-server 200 '{"runId":"run-123","statusCode":202}'
+snouty launch -w basic_test --duration 30 --filter_logs_matching debug
+stderr '"antithesis.duration": "30"'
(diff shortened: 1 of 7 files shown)


carlsverre:
@agupps what happens if the tenant is not v59 or above? Does the run fail with a useful error message like "contact support to upgrade" or does the webhook endpoint reject?

agupps:
> @agupps what happens if the tenant is not v59 or above? Does the run fail with a useful error message like "contact support to upgrade" or does the webhook endpoint reject?

@carlsverre On the product side, it'll accept the parameter and then nothing will actually get filtered. I can make snouty, if --filter-logs-matching is passed in, it can get the Antithesis version via API (unless you have a better way to get it) and then see if its <v59 and print out a validation error like (--filter-logs-matching not supported for this version). Or do you prefer something else.

carlsverre:
A pre-flight check makes sense to me. But perhaps as a separate PR. I'll take this one, as I also want to make the version subcommand support checking the tenant version.

PR #204, opened 2026-08-05
Refresh openapi.json to tenant release 58.11 and pin the spec workarounds

carlsverre:
This refreshes `src/openapi.json` from the orbitinghail tenant (release 58.11; the v0 and v1 spec routes serve one identical document) and adjusts build.rs to the new spec.

Spec changes:

- New `POST /runs/{run_id}/events/search` endpoint with the event-set DSL.
- New `Stream_Error` schema, documented as a possible NDJSON line on the `build_logs`, `events`, and `logs` streams.
- The events `limit` maximum rises from 999 to 1000.

The new spec needs two build.rs transforms. First, the spec marks `Search_Request` and `Search_Count_Response` with `additionalProperties: false`. progenitor/typify turns this into `#[serde(deny_unknown_fields)]`, which makes a generated client reject any response field the spec does not list. build.rs strips the constraint before generation. This restores the transform that we removed when release 58.6 shipped a clean spec. Second, the search endpoint description contains an untagged Markdown code fence around event-DSL examples. progenitor copies descriptions into doc comments, and rustdoc compiles untagged fenced blocks as Rust doctests. build.rs tags untagged fences as ```text.

Both transforms assert an exact pinned occurrence count. When a spec refresh moves a count in either direction, the build fails with a message that says to remind the API team and update the pin.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,6 +16,15 @@ fn main() {
     generate_api_client(Path::new(&out_dir));
 }
 
+/// How many `"additionalProperties": false` occurrences the vendored spec
+/// carries (tenant release 58.11: `Search_Request`, `Search_Count_Response`).
+const EXPECTED_ADDITIONAL_PROPERTIES_FALSE: usize = 2;
+
+/// How many untagged Markdown code fences the vendored spec's descriptions
+/// carry (tenant release 58.11: the example block in the events-search
+/// endpoint description).
+const EXPECTED_UNTAGGED_CODE_FENCES: usize = 1;
+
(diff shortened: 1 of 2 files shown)


carlsverre:
Superseded by #205, which absorbs this PR: it carries the same build.rs shape — both spec transforms pinned to exact occurrence counts — on top of a newer spec refresh (release 60.0 vs 58.11). Release 60.0 drops the untagged code fence, so the fence pin there is zero with the tagging transform still in place. #205 also adds scripts/refresh-openapi.sh for future refreshes.

PR #205, opened 2026-08-05
Refresh openapi.json to release 60.0, pin the spec transforms, and fail streaming commands on Stream_Error

carlsverre:
This PR refreshes the vendored OpenAPI spec and makes snouty handle the `Stream_Error` line the new spec documents. It absorbs #204, which refreshed to release 58.11; the spec here is release 60.0, so this PR's `openapi.json` is ahead of #204's.

## Spec refresh (tenant release 60.0)

`src/openapi.json` is refreshed from the tenant (the v0 and v1 spec routes serve one identical document). The spec adds two endpoints — `POST /runs/{run_id}/events/search` and `POST /runs/{run_id}/execute_command` — which snouty does not adopt yet, and a `Stream_Error` schema: every streaming NDJSON response (logs, events, build logs) may now end with an `{"error": "..."}` line when the server fails after it has already committed to a `200 OK`.

A new `scripts/refresh-openapi.py` (run via uv, like the other Python helper scripts) performs the refresh. It resolves auth and tenant from the environment in the same order snouty does (`ANTITHESIS_API_KEY` or the ApiKey profile in `~/.config/snouty/credentials.toml`; `ANTITHESIS_BASE_URL`, `ANTITHESIS_TENANT`, or the tenant in `settings.toml`), and passes the download through `jq .` before writing, so the vendored file carries one consistent formatting and refresh diffs stay free of formatting noise. The vendored spec is reformatted with the same filter (semantically unchanged). A `--version vN` flag (default `v1`) selects the spec route to fetch. The script also writes `src/openapi.provenance.json` — the tenant release version and latest API version from `GET /api/version`, plus the spec route fetched — so the vendored spec's provenance is on record.

## Pinned spec transform (from #204)

build.rs adopts #204's shape: the spec workaround asserts an exact pinned occurrence count, so what we expect of the spec at compile time is explicit, and a refresh that moves the count in either direction fails the build with an ACTION message — remind the API team about the defect, then update the pin.

`EXPECTED_ADDITIONAL_PROPERTIES_FALSE = 2`: `Search_Request` and `Search_Count_Response` mark `additionalProperties: false`, which progenitor/typify turns into `#[serde(deny_unknown_fields)]` — a client that rejects any response field the spec does not list. build.rs strips the constraint before generation (the transform release 58.6 had made unnecessary).

(#204 also carried a transform that tags untagged Markdown code fences in descriptions, which rustdoc would compile as failing doctests. Release 60.0 carries no fenced blocks, so per review that transform is not brought over; it lives in #204/git history if a fence returns.)

## Fail streaming commands on a mid-stream `Stream_Error`

snouty treated the error line as data, and the failure modes ranged from bad to invisible:

- `runs build-logs` (human) rendered it as a blank ` [out] ` line — the message was dropped entirely.
- `runs events` filtered it out through the needle matcher and could then report `No events matched`.
- `runs logs` (human) printed it disguised as a log record with blank vtime/source/stream columns; `--json` injected `active_faults` into it and emitted it as data.
- In every mode the command exited 0 over a truncated stream, and nothing reached stderr.

The fix turns the line into the command's failure: the server's message goes to stderr with a note that prior output may be incomplete, and the exit code is non-zero. The error line reaches stdout in no mode, including the raw NDJSON passthroughs — it is the server aborting the stream, not data. Output produced before the error still prints; `runs events` renders the rows it buffered before it propagates the failure. Layering: `split_stream_lines` is the pure byte-to-lines splitter with no JSON awareness, and each NDJSON wrapper detects the error at its own decode pass — `stream_ndjson_lines` inspects the record it already parsed, and `stream_raw_lines` (the verbatim-passthrough wrapper) takes one decode pass whose only purpose is the detection.

A data record can carry a top-level `error` field of its own (an event's payload is arbitrary workload JSON), so a line only counts as a `Stream_Error` when it lacks the required envelope of both data shapes: `moment` on log/event records and `timestamp` on build log lines.

Tests: a `run-stream-error` mock fixture ends each of its streams with the error line; specs cover all seven command modes, and unit tests pin the `Stream_Error`-versus-data discrimination.


build.rs
```diff
@@ -16,6 +16,15 @@ fn main() {
...
 }
 
+/// How many `"additionalProperties": false` occurrences the vendored spec
+/// carries (tenant release 60.0: `Search_Request`, `Search_Count_Response`).
+const EXPECTED_ADDITIONAL_PROPERTIES_FALSE: usize = 2;
+
+/// How many untagged Markdown code fences the vendored spec's descriptions
+/// carry (tenant release 60.0: none — release 58.11's event-DSL example block
+/// in the events-search endpoint description is gone).
+const EXPECTED_UNTAGGED_CODE_FENCES: usize = 0;
```
carlsverre: remove the untagged code fences transform - if it comes back I'll re-add it.

carlsverre: Removed in bf23577. The 60.0 spec carries no fenced blocks, so the transform guarded nothing; it stays in git history for re-adding.

scripts/refresh-openapi.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def load_toml(path: Path) -> dict:
+    if not path.is_file():
+        return {}
+    with path.open("rb") as f:
+        return tomllib.load(f)
+
+
+def resolve_api_key() -> str:
```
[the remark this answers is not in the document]
carlsverre: Fixed in 882fa29: `fail()` now builds a `SystemExit` and every call site raises it, so the control flow is explicit. Behavior is unchanged (a raised `SystemExit` with a string prints it to stderr and exits 1).

PR #206, opened 2026-08-05
Add `snouty runs search` (gated) and rebuild `runs events` on top of it

carlsverre:
This exposes the events-search API (`POST /runs/{run_id}/events/search`) as `snouty runs search RUN_ID QUERY`, behind the `runs-search` unstable feature, and rebuilds `runs events` on top of the same stream. It builds on #205's spec refresh and Stream_Error handling.

**The feature gate.** The events-search API does not honor its documented contract yet on current tenants (observed on releases 58.11 through 60.1): the server ignores the request's `limit` on runs with a live stream, and holds non-streaming connections to live runs open instead of closing them after the current matching set. `runs search` is therefore gated behind `SNOUTY_UNSTABLE_FEATURES=runs-search` until the server fixes both. snouty enforces the limit client-side — the flag's value, or the server default of 50 when the request named none — and that cap is the only early exit: fewer events on a silent open connection means keep waiting, never "assume complete". The one uncapped shape is `-f/--follow` without an explicit limit: the request carries no `limit` field at all (build.rs strips the schema's `default: 50` so the generated field is omittable — the server is starting to honor `limit` together with `is_streaming`, and a baked-in 50 would cut an unbounded follow off), and the client runs uncapped too.

**One stream type.** Every Antithesis streaming endpoint answers JSONL, so every streaming api method — logs, build logs, events, search, execute-command — now returns a `JsonlStream` (`src/jsonl.rs`): the HTTP byte stream split into lines and parsed with serde as early as possible. serde_json's `preserve_order` + `float_roundtrip` make the parsed Value a faithful currency, so raw modes round-trip Values instead of echoing bytes. The line-splitting helpers in `runs.rs` collapsed into consumers of the shared type.

**Backends resolve up front.** `runs events` picks its backend from the feature flag, in the command and nowhere else: with `runs-search` on, the events-search endpoint ANDs every needle server-side against the whole raw event JSON (a JS `filter`, so structured events with no `output_text` — assertions, composer and fault records — stay findable); with it off, the GET events endpoint takes exactly one substring, and several needles fail up front with the fix. `runs search` always uses events-search. With the feature on, the tenant is assumed to serve the endpoint — a 404 simply means the run — and `snouty doctor` owns the assumption: it fails when the feature is enabled and the tenant release predates 58.11.

**One dumb output path.** From the stream on, the two commands are identical: each matching event prints as its JSON, one unbuffered line per event with `moment.vtime` normalized to an exact number, flushed per row. #222 layers the human rendering on this. A `{"error": ...}` line is the server's mid-stream failure signal on every fixed-schema stream and fails the command — except on a `runs search` stream, where a user-written pipeline can legitimately reshape a row into exactly that shape (`map(ev => ({error: ev.output_text}))`), so it prints as data.

**Behavior changes to `runs events`** (also in the CHANGELOG):

- All client-side re-filtering is removed. Filtering the output of a server-side filter is never valid: whatever the server returns now prints.
- The buffered aligned table is gone; output is one JSON line per event.
- Multiple `--match` needles require the search backend; without the feature the command fails up front and names the fix.
- `--limit` is `NonZeroU64` end to end on both commands: `--limit 0` is a clap error that names the flag.

**Supporting changes.** The `runs search` help documents every DSL verb in one line each and carries ten copy-paste-ready query snippets (substring/regex/whole-event matching, per-container stderr, vtime windows, assertion queries by id and by type, and a crash-annotated-with-last-fault temporal query), each validated with `--check` and executed against staging. `event_set_dsl` keeps the verb list and the needles-to-`filter()` builder; the quoted-literal extractor the mock uses lives in the mock server code. build.rs drops the `application/json` variant from the search operation's 200 response so the generated client exposes the NDJSON stream, and strips the `limit` default as above (both asserted, so a spec refresh that fixes them upstream fails the build as a reminder). The body's `count_only` switch is deliberately not exposed: the API team is moving the count to a separate endpoint.

Verified against the orbitinghail tenant (release 60.1): the staging spec suite; the gate; search limits and `--check` on valid and invalid queries; every help snippet executed live; events single- and multi-needle on both backends; `run not found` on bad run ids; and the omitted-limit request body pinned by a unit test.


src/runs.rs
```diff
@@ -1343,6 +1361,268 @@ fn haystack_matches_all_needles(haystack: &str, lowered_needles: &[String]) -> b
...
+        .cloned()
+        .unwrap_or_default();
+    eprintln!(
+        "Note: only \"{server_query}\" is matched on the server (which returns a capped \
+         subset of events); the other terms filter that subset locally, so some matching \
+         events may not appear. Search a single, more specific term to be exhaustive, or \
+         raise the cap with --limit."
+    );
+    match api.search_run_events(run_id, &server_query, limit).await {
+        Ok(stream) => Ok(stream),
```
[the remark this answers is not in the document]
carlsverre: Fixed in 5edcaaa. The server-side prefilter now matches each needle against the whole raw event JSON (`JSON.stringify(ev).toLowerCase()`), so assertion, composer, and fault events with no `output_text` stay findable; the client pass over the rendered haystack keeps displayed matches exact. Verified live: `runs events RUN partition jammed` returns fault events that carry no `output_text` (both needles exist only in the structured payload). The residual gap — a needle that exists only in the rendering, e.g. `must-hit` for the raw `must_hit` — is documented on `needles_to_dsl_filter` and in the PR body; the legacy path had the same gap for its one server-matched needle.

src/runs.rs
```diff
@@ -1343,6 +1361,268 @@ fn haystack_matches_all_needles(haystack: &str, lowered_needles: &[String]) -> b
...
+        // (with and without `is_streaming`), so rows must not sit in the
+        // buffer waiting for an EOF that may never come.
+        stdout.flush()?;
+        // The server is supposed to end the stream at `limit`, but a live
+        // run's stream ignores it (observed on tenant release 58.11), so the
+        // documented contract is enforced client-side too.
+        match modes.limit {
+            Some(cap) if seen >= cap => Ok(ControlFlow::Break(())),
+            _ => Ok(ControlFlow::Continue(())),
+        }
```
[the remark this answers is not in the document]
carlsverre: Fixed in 5edcaaa. `cmd_runs_search` now enforces the effective limit client-side: the `-n` value, or 50 — the server default that the request body always names. It applies in `--follow` mode too, matching the documented contract (the server terminates at `limit` whether or not `is_streaming` is set). Verified live: a no-flag `runs search` against an in-progress run printed 50 rows and exited in 0.6s.

src/cli.rs
```diff
@@ -658,6 +658,45 @@ and VTIME into `runs logs` to see the surrounding logs."#
...
+
+        /// Keep the connection open and print new matches as they arrive
+        #[arg(short = 'f', long)]
+        follow: bool,
+
+        /// Check the query's syntax without running it
+        #[arg(long, conflicts_with = "follow")]
+        check: bool,
+    },
 }
```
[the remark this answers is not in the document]
carlsverre: This review ran against the pre-rebase commit. The current commit (5edcaaa) extends `specs/runs.txt` with coverage for every `runs search` mode — fetch, `--json`, `--limit`, empty result, `--count`, `--check` valid and invalid, conflicting switches, unknown run, and the `run-stream-error` fixture — plus a pin that the multi-needle caveat note no longer prints for terminal runs. The mock API server implements the endpoint to make those specs run.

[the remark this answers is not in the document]
carlsverre: Same as the sibling comment: stale — the current commit (5edcaaa) carries the specs in `specs/runs.txt` and the mock-server support for them.

carlsverre:
@claude review

src/event_set_dsl.rs
```diff
@@ -0,0 +1,125 @@
...
+        .collect::<Vec<_>>()
+        .join(" && ");
+    format!("filter(ev => ((h) => {clauses})(JSON.stringify(ev).toLowerCase()))")
+}
+
+/// Every double-quoted string literal in `query`, with `\"` and `\\` escapes
+/// resolved. The mock API server uses this to "interpret" a query without a
+/// DSL engine: requiring each literal somewhere in an event covers
+/// `contains({...})` needles and [`substring_filter`] pipelines alike.
+pub fn extract_quoted_literals(query: &str) -> Vec<String> {
```
carlsverre: move this into the mock server code - so it's clear that it's testonly

src/api.rs
```diff
@@ -457,22 +491,56 @@ impl AntithesisApi {
...
+        query: &str,
+        opts: EventSearchOptions,
+    ) -> Result<ByteStream> {
+        // `count_only` is left to the generated default (false): snouty does
+        // not expose it (see `EventSearchOptions`).
+        let mut body = generated::types::SearchRequest::builder()
+            .query(query)
+            .is_streaming(opts.is_streaming)
+            .validate_only(opts.validate_only);
+        // The generated body type defaults `limit` to the server's default
```
carlsverre: this is a bug - if is_streaming is true and the user did not provide an explicit limit then we must not set any limit on the request. if progenator is doing that for us then we should change the openapi spec accordingly. the api server team is going to start respecting limit + stream soon

carlsverre (commented):
lets address a structural problem with this code first. currently dispatch resolves which search backend is enabled. this is a bad design as it requires plumbing unrelated concepts down through many layers.

let's use this design instead:

First let's define a standard type across the codebase for jsonl streams. All streaming apis from Antithesis are jsonl streams. This includes logs, build_logs, events, search, and execute_command. plus future endpoints. We should parse JSONL streams with serde as early as possible.

Looking at this PR I think the attempt to hide events and search behind an API ended up creating a complicated and hard to grok mess of code. This is what we actually want:

In the events subcommand:
```
let stream = if search feature enabled {
  let query = query_from_needles(cli.matches)
  api.search(query, follow=false, limit=cli.limit, check=false)?
} else {
  bail if matches.len != 1
  api.events(matches[0], limit=cli.limit)?
}
```

The search subcommand would just always use api.search to generate the stream.

Once you have the stream - everything from there is completely consistent between the commands. A lot of this is going to be handled in PR 222 so keep this simple. ideally just dump the stream to stdout. let PR 222 render the output nicely and handle various modes.

I suspect this should allow you to dramatically simply this PR. the raw flag effectively just means we round trip the raw serde json Values back to stdout without normalization. otherwise we normalize and render the output either as json or a pretty object (which PR 222 will do, so do something simple/dumb for now)

Also you can assume that if the search feature is enabled then the tenant has it. which will simplify error handling in events. But Doctor can properly provdide a nice error if the search feature is enabled && the release version is not recent enough

src/runs/event_search.rs
```diff
@@ -0,0 +1,178 @@
...
+/// an unhelpful "Resource not found". A 400 is the server rejecting the
+/// request — usually the query, but an out-of-range `limit` answers 400 too,
+/// so the note must not blame the query alone (the server's own message,
+/// shown above the note, says which it was).
+pub(super) fn explain_search_error(
+    run_id: &str,
+    err: color_eyre::eyre::Report,
+) -> color_eyre::eyre::Report {
+    match api_error_status(&err) {
+        Some(404) => user_error(format!("run not found: {run_id}")),
```
carlsverre: for now, add a suggestion to this error that says tenant versions < X do not support search

src/api.rs
```diff
@@ -57,6 +60,29 @@ pub struct ApiVersion {
...
+/// tenants ignore the switch anyway (observed on releases 58.11 and 60.0).
+#[derive(Debug, Clone, Copy, Default)]
+pub struct EventSearchOptions {
+    pub is_streaming: bool,
+    pub validate_only: bool,
+    /// Cap on returned events (server-validated range 1..=999, default 50).
+    /// `NonZeroU64` matches the generated `Search_Request` body's field, so a
+    /// zero can never reach the request builder.
+    pub limit: Option<NonZeroU64>,
+}
```
carlsverre: make this an enum `{ Validate, Query { stream, limit } }`

src/api.rs
```diff
@@ -457,18 +483,54 @@ impl AntithesisApi {
...
+    /// lines, and the `validate_only` answer as an empty body. build.rs drops
+    /// the operation's `application/json` response variant so the generated
+    /// method exposes the stream (see `untype_search_count_response` there).
+    pub async fn search_run_events_query(
+        &self,
+        run_id: &str,
+        query: &str,
+        opts: EventSearchOptions,
+    ) -> Result<JsonlStream> {
+        // `count_only` is left to the generated default (false): snouty does
```
carlsverre: modify build.rs to strip the count_only param entirely from the type (unless the server rejects the request if it's missing)

src/api.rs
```diff
@@ -57,6 +60,29 @@ pub struct ApiVersion {
...
+    /// Cap on returned events (server-validated range 1..=999, default 50).
+    /// `NonZeroU64` matches the generated `Search_Request` body's field, so a
+    /// zero can never reach the request builder.
+    pub limit: Option<NonZeroU64>,
+}
+
+/// The server's default for the search `limit`, applied when the request
+/// names none. A non-streaming caller that enforces the limit client-side
+/// caps at this value when no explicit limit was given.
+pub const SEARCH_DEFAULT_LIMIT: NonZeroU64 = NonZeroU64::new(50).unwrap();
```
carlsverre: this looks unused

src/doctor.rs
```diff
@@ -370,6 +370,52 @@ fn print_settings(settings: &[Setting]) {
...
+
+/// With the `runs-search` unstable feature enabled, `runs events` and
+/// `runs search` assume the tenant serves the events-search API instead of
+/// probing for it — this check is where that assumption gets verified. Only
+/// a confidently-known gap reports: the feature off, an unreachable version
+/// endpoint, or an unparsable release version all say nothing (the check
+/// would guess). Pure over the probe result so it can be unit-tested without
+/// the network.
+fn runs_search_release_check(
+    result: &std::result::Result<ApiVersion, VersionError>,
```
carlsverre: resolve the result elsewhere. do not pass Results into functions

src/doctor.rs
```diff
@@ -370,6 +370,52 @@ fn print_settings(settings: &[Setting]) {
     }
 }
 
+/// A tenant release version's comparable form: the leading `major.minor`
+/// pair of a string like `"60.1"`. `None` when the string does not lead with
+/// two integers.
+fn parse_release(version: &str) -> Option<(u64, u64)> {
```
carlsverre: do this when we create the ApiVersion type. users of ApiVersion should not have to parse. Always parse eagerly.

src/doctor.rs
```diff
@@ -370,6 +370,52 @@ fn print_settings(settings: &[Setting]) {
...
+    let release = parse_release(&version.release_version)?;
+    if release >= RUNS_SEARCH_MIN_RELEASE {
+        return None;
+    }
+    Some(
+        Check::fail("runs-search", "tenant serves the events-search API").note(
+            Level::Error,
+            format!(
+                "the `runs-search` unstable feature is enabled, but tenant release {} \
+                 predates the events-search API (added in 58.11) — `runs events` and \
```
carlsverre: 58.11 should come from the constant

src/doctor.rs
```diff
@@ -370,6 +370,52 @@ fn print_settings(settings: &[Setting]) {
...
+                 `runs search` will fail",
+                version.release_version
```
carlsverre: suggest that the user removes `runs-search` from SNOUTY_UNSTABLE_FEATURES

src/event_set_dsl.rs
```diff
@@ -0,0 +1,82 @@
...
+/// through an immediately-invoked arrow function, so the server stringifies
+/// and lowercases each event once, not once per needle (the form is accepted
+/// by the server's validator; verified on tenant release 60.1). The raw JSON
+/// — not snouty's rendered form — is the only haystack: a needle that exists
+/// only in the rendering (e.g. the assertion summary's `must-hit` marker for
+/// the raw `must_hit` field) finds nothing, and conversely a match inside a
+/// field the rendering elides still returns its event. That asymmetry is the
+/// accepted cost of filtering entirely server-side. Each needle is embedded
+/// as a JSON string literal (also a valid JS string literal), so escaping is
+/// serde's problem, not string surgery here.
```
carlsverre: Simplify this comment. Don't explain implementation details in comments unless they affect the function's preconditions

src/jsonl.rs
```diff
@@ -0,0 +1,164 @@
...
+//! JSON object per line.
+//!
+//! Every streaming endpoint — logs, build logs, events, event search,
+//! execute-command, and whatever comes next — answers with this shape, so
+//! every one of them comes out of [`crate::api`] as a [`JsonlStream`]: the
+//! HTTP byte stream split into lines and parsed with serde as early as
+//! possible. Consumers see [`JsonlLine`]s, never raw bytes.
+//!
+//! Parsing this early costs nothing in fidelity: serde_json is built with
+//! `preserve_order` and `float_roundtrip`, so a [`Value`] serialized back out
```
carlsverre: drop preserve_order if it's costly perf wise. I don't care about preserving the order of keys.

src/jsonl.rs
```diff
@@ -0,0 +1,164 @@
...
+            inner: stream.boxed(),
+            buf: Vec::new(),
+            done: false,
+        }
+    }
+
+    /// The next line of the stream, or `None` when the stream is over.
+    /// Empty lines are skipped; a final line without a trailing newline
+    /// still counts.
+    pub async fn next_line(&mut self) -> Result<Option<JsonlLine>> {
```
carlsverre: 1. Rename JsonlStream to JsonStream
2. Replace `next_line` with a full Stream impl that emits `serde_json::Value` as it's value
3. just fail the stream with the corrupted raw line in the error if you run into an invalid JSONL line rather than JsonlLine::Raw

src/runs.rs
```diff
@@ -1494,137 +1521,53 @@ async fn cmd_runs_events(
...
-            }
-            NdjsonLine::Raw(line) => (
-                line.to_string(),
-                vec![String::new(), String::new(), String::new(), sanitize(line)],
-            ),
+    // The backend is the feature flag's call, made here and nowhere else.
+    // With `runs-search` on, the events-search endpoint ANDs every needle
+    // server-side against the whole raw event JSON (the tenant is assumed to
+    // serve it; `snouty doctor` reports a tenant release that is too old).
+    // Without it, the GET events endpoint takes exactly one substring.
```
carlsverre: delete comment

carlsverre (commented):
Once you make JsonlStream a real futures::Stream you should be able to simplify a ton of stream processing code with things like Stream::map to do things like normalization or conversion into raw values. We have a ton of functions like stream_ndjson_lines_capped which just seem like poor mans implementations of Stream builtins. Similarly cap can simply be Stream::take.

carlsverre:
@claude review

src/jsonl.rs
```diff
@@ -0,0 +1,156 @@
...
+    serde_json::from_str(line).map_err(|e| {
+        eyre!("invalid JSON in the response stream: {e}").note(format!(
+            "the corrupted line: {}",
+            crate::render::sanitize(line)
+        ))
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
carlsverre, line 91: a hegel test that constructs a random series of json values, serializes them into a jsonl buffer, then randomly splits the buffer into chunks and verifies the whole thing round trips back to the same series of json values at the end (after passing through JsonStream) would be a great test

src/jsonl.rs
```diff
@@ -136,6 +136,82 @@ mod tests {
...
+
+        // ... split into arbitrary chunks, so line boundaries land anywhere.
+        let mut chunks: Vec<reqwest::Result<Bytes>> = Vec::new();
+        let mut rest = buf.as_slice();
+        while !rest.is_empty() {
+            let n = tc.draw(
+                generators::integers::<usize>()
+                    .min_value(1)
+                    .max_value(rest.len()),
+            );
```
carlsverre, line 202: can a response body chunk ever be empty? if so min_value should be 0 here.

PR #208, opened 2026-08-06
Add snouty runs exec for the execute-command API

carlsverre:
This PR adds the `snouty runs exec` command. The command runs a bash script in a test run's live session, at a moment. It uses the `execute_command` API, which #205 added to the vendored spec.

```
export SNOUTY_UNSTABLE_FEATURES=runs-exec
snouty runs exec <RUN_ID> <INPUT_HASH> <VTIME> 'uname -a'
echo 'ps aux' | snouty runs exec <RUN_ID> <INPUT_HASH> <VTIME>
```

I tested the API against the live orbitinghail tenant. The command needs a live session. A run keeps one while it runs, and for a while after it completes. A run whose session has ended answers with 400. The length of that window is not documented: a run that completed 48 minutes earlier still executed a script, and runs that completed 17 hours and 1 day earlier did not. The script runs on a new branch of the multiverse, and the response moments carry a new `input_hash`. Thus the script does not disturb the running test.

The API is unstable, and most tenants cannot serve it. Thus the command ships in the off state. To enable a feature, put its id in `SNOUTY_UNSTABLE_FEATURES`. The variable holds a list of ids, separated by commas. The name says "unstable" because that is the promise: a feature behind this gate can change its behavior, its flags, or its id, or go away, in any release.

The README does not document the variable or the command. An unstable feature is not part of the public surface, so it stays out of the user-facing documentation until it is stable. `snouty runs exec --help` and `snouty doctor` describe it for anyone who enables it.

`SNOUTY_UNSTABLE_FEATURES` is an environment variable, not a settings key. The gate selects which subcommands the parser has. Thus snouty must know the gate before it parses the command line. It cannot read a settings file that early, because `--settings` and `--profile` give the location of that file. clap's `hide` attribute accepts an expression, so the gate needs no second parse. `snouty doctor` shows the features that are on.

While the feature is off, `--help` does not show `runs exec`. An invocation of the command fails with clap's unrecognized-subcommand error and exit code 2.

The gate hides the command, but it does not remove the command. Two effects follow, and `specs/features.txt` pins both. First, `runs exec --help` still prints the help. That help names the feature and tells the user how to enable it. Second, the bash and zsh completions still offer `exec`, because clap_complete puts hidden subcommands in the candidate list.

`specs/runs_exec.txt` pins the contract of the command:

- The script comes from the `SCRIPT` argument. If you omit the argument, the command reads the script from stdin. A pipe, a redirect, and a heredoc all operate. This is the only way to supply the script on stdin. The limit for stdin is 2 MiB, and the command reports a script above that limit by its size. It rejects a script that is not valid UTF-8 with its own message. The command refuses to read from an interactive terminal. Thus a bare invocation at a prompt gives an explanation, and does not look like a hang.
- The command sends each `output` event to the related stream. It sends stdout lines to snouty's stdout, and stderr lines to snouty's stderr. Only the script's output goes to stdout. Thus `snouty runs exec ... | jq` operates correctly.
- After an `exited` event, a trailer on stderr gives the end moment of the branch. The trailer has the form `end moment: <hash> <vtime>`, in the same order as the positional arguments. Use these values to run the next command from that moment.
- An `exited` event with code 0 is a success. Each other terminal state fails the command with exit code 1: a non-zero code, which snouty reports but does not forward; a `timed_out` event; or a stream that ends with no terminal event.
- clap parses `VTIME` into the `VTime` type. Thus snouty rejects a bad value before it calls the API.
- `--timeout <seconds>` becomes the `timeout_seconds` field. The default is 30, which is also the API default. The API gives a minimum of 0 and no maximum. snouty sets the minimum to 1, because a timeout of 0 seconds can only time out.
- `--json` sends the NDJSON through without change, but it makes each `moment` and `end_moment` vtime an exact JSON number. This agrees with `runs logs --json`. In this mode snouty prints no trailer.
- A `Stream_Error` line becomes the failure of the command. This is the behavior from #205.
- snouty sends the request for any run id. It pins no minimum tenant release, unlike the run properties and build logs resources. The multiverse debugger has not shipped, so the release that first serves this endpoint is not settled, and a pinned floor would reject runs the server can in fact execute against. The server's own answer is the authority.
- The server answers a run whose session has ended with a rough 400 message ("minting session token failed"). snouty shows this message without change. The API team improves the message on the server.

The PR also makes each command-line vtime use the `VTime` type. `runs logs` takes `VTIME` and `--begin-vtime` as typed values. `debug --vtime` is also typed. A debug moment arrives by three routes: the typed flags, a `Moment.from` on stdin, and raw JSON on stdin. `Params::normalize_debug_vtime` normalizes the vtime one time, after these routes merge. Thus all three routes send the same text for the same moment.

Changes:

- `specs/runs_exec.txt` and `specs/features.txt` are the two new specs. A `[staging] skip` line gates `runs_exec.txt`, like `debug.txt`, because each successful invocation runs a script in a live session.
- `specs/runs.txt` and `specs/debug.txt` gain cases for the typed vtimes. Each one drives a bad value and expects a local error.
- `src/features.rs` holds the `Feature` type and the reader for `SNOUTY_UNSTABLE_FEATURES`. It keeps an unknown id instead of a rejection. Thus one exported variable stays usable across snouty versions.
- `src/cli.rs` holds the `runs exec` subcommand, its `hide` expression, and the refusal of a gated-off command.
- `src/api.rs` holds `execute_command()`.
- `src/runs.rs` holds the streaming command, the script resolution, and the end-moment trailer.
- `src/params.rs` holds `normalize_debug_vtime`.
- `src/moment.rs` drops its own vtime rule, because `normalize_debug_vtime` now owns that rule for every route. `moment::parse` only turns numbers into strings.
- `src/doctor.rs` adds a `features` row to the resolved-configuration table. The row appears only when a feature is on.
- `src/testutils.rs` reads request bodies up to Content-Length, because one `read` call raced on POST bodies. It also serves an `execute_command` route. The fixtures of that route copy the behavior of the live API.
- `tests/cli_general.rs` expects the normalized vtime text in the debug params.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -353,6 +353,32 @@ impl AntithesisApi {
         }
     }
 
+    /// Execute a bash script in the run's live session, starting at `moment`.
+    /// Returns the NDJSON response stream: `output` events, then a terminal
+    /// `exited` or `timed_out` event.
+    pub async fn execute_command(
+        &self,
+        run_id: &str,
+        moment: Moment,
+        script: &str,
+        timeout_seconds: u64,
+    ) -> Result<ByteStream> {
+        ensure_resource_supported(run_id, MIN_EXEC_VERSION, "command execution")?;
+        let body = generated::types::ExecuteCommandRequest {
+            moment,
+            script: script.to_string(),
+            timeout_seconds,
+            // A server-side testing knob (routes output through the tenant
+            // coordinator); snouty always wants the live session's output.
+            use_otis: false,
+        };
+        let request = self.client.execute_command().run_id(run_id).body(body);
+        match request.send().await {
+            Ok(response) => Ok(response.into_inner()),
+            Err(err) => Err(format_api_client_error(err).await),
+        }
+    }
+
     pub fn stream_run_properties(
         &self,
         run_id: &str,
@@ -897,6 +923,10 @@ const MIN_PROPERTIES_VERSION: u32 = 52;
 /// properties arrived in v52, while logs and events are served for every
 /// version we can produce, so neither needs a guard.
 const MIN_BUILD_LOGS_VERSION: u32 = 54;
+
+/// The tenant version that first served the execute-command resource. Runs
+/// created on older tenants 404 on `/runs/{run_id}/execute_command`.
+const MIN_EXEC_VERSION: u32 = 60;
 
 /// Run IDs encode the tenant version that produced them as their second
 /// dash-delimited field — e.g. the `40` in
--- a/src/api.rs
+++ b/src/api.rs
@@ -372,7 +372,6 @@ impl AntithesisApi {
         script: String,
         timeout_seconds: u64,
     ) -> Result<ByteStream> {
-        ensure_resource_supported(run_id, MIN_EXEC_VERSION, "command execution")?;
         let body = generated::types::ExecuteCommandRequest {
             moment,
             script,
@@ -929,14 +928,16 @@ const MIN_PROPERTIES_VERSION: u32 = 52;
 /// created on older tenants 404 on `/runs/{run_id}/build_logs`.
 const MIN_BUILD_LOGS_VERSION: u32 = 54;
 
-/// The tenant version that first served the execute-command resource. Runs
-/// created on older tenants 404 on `/runs/{run_id}/execute_command`.
-const MIN_EXEC_VERSION: u32 = 60;
-
 // Every nested run resource with a version cutoff has a `MIN_*_VERSION` const
 // above, enforced by `ensure_resource_supported`. Logs and events have no
 // cutoff — they are served for every version we can produce — so neither
-// needs a guard. A new gated resource belongs in this group.
+// needs a guard.
+//
+// Execute-command has no cutoff either, on purpose. The multiverse debugger
+// has not shipped, so the release that first serves the endpoint is not
+// settled, and a pinned floor here would reject runs the server can in fact
+// execute against. The feature is unstable, so the server's own answer is the
+// authority. Add a const here once the release is known.
 
 /// Run IDs encode the tenant version that produced them as their second
 /// dash-delimited field — e.g. the `40` in
--- a/src/api.rs
+++ b/src/api.rs
@@ -326,9 +326,9 @@ impl AntithesisApi {
         }
     }
 
-    /// The endpoint takes each vtime as a decimal-seconds query parameter, so
-    /// the typed values are rendered with [`VTime`]'s exact `Display` — the
-    /// same text the value was parsed from, for anything snouty printed.
+    /// The endpoint takes each vtime as a decimal-seconds query parameter. The
+    /// generated setters accept anything that converts to `String`, and
+    /// [`VTime`]'s conversion is its exact `Display` text.
     pub async fn get_run_logs(
         &self,
         run_id: &str,
@@ -342,12 +342,12 @@ impl AntithesisApi {
             .get_run_logs()
             .run_id(run_id)
             .input_hash(input_hash)
-            .vtime(vtime.to_string());
+            .vtime(vtime);
         if let Some(v) = begin_input_hash {
             request = request.begin_input_hash(v);
         }
         if let Some(v) = begin_vtime {
-            request = request.begin_vtime(v.to_string());
+            request = request.begin_vtime(v);
         }
 
         match request.send().await {
@@ -372,13 +372,12 @@ impl AntithesisApi {
         script: String,
         timeout_seconds: u64,
     ) -> Result<ByteStream> {
+        // No `use_otis`: `build.rs` drops that field from the generated type,
+        // so snouty says nothing about it and the server applies its default.
         let body = generated::types::ExecuteCommandRequest {
             moment,
             script,
             timeout_seconds,
-            // A server-side testing knob (routes output through the tenant
-            // coordinator); snouty always wants the live session's output.
-            use_otis: false,
         };
         let request = self.client.execute_command().run_id(run_id).body(body);
         match request.send().await {
@@ -927,17 +926,6 @@ const MIN_PROPERTIES_VERSION: u32 = 52;
 /// The tenant version that first served the run build logs resource. Runs
 /// created on older tenants 404 on `/runs/{run_id}/build_logs`.
 const MIN_BUILD_LOGS_VERSION: u32 = 54;
-
-// Every nested run resource with a version cutoff has a `MIN_*_VERSION` const
-// above, enforced by `ensure_resource_supported`. Logs and events have no
-// cutoff — they are served for every version we can produce — so neither
-// needs a guard.
-//
-// Execute-command has no cutoff either, on purpose. The multiverse debugger
-// has not shipped, so the release that first serves the endpoint is not
-// settled, and a pinned floor here would reject runs the server can in fact
-// execute against. The feature is unstable, so the server's own answer is the
-// authority. Add a const here once the release is known.
 
 /// Run IDs encode the tenant version that produced them as their second
 /// dash-delimited field — e.g. the `40` in
--- a/src/api.rs
+++ b/src/api.rs
@@ -356,17 +356,23 @@ impl AntithesisApi {
     /// Execute a bash script in the run's live session, starting at `moment`.
     /// Returns the NDJSON response stream: `output` events, then a terminal
     /// `exited` or `timed_out` event.
+    ///
+    /// The request body carries `moment.vtime` as a JSON number ([`VTime`]'s
+    /// wire form), where the spec documents a string. The server accepts both
+    /// — verified against the live API (orbitinghail, release 60.0) with this
+    /// exact request path — and the number form keeps the moment value-exact
+    /// end to end instead of round-tripping through a second text form.
     pub async fn execute_command(
         &self,
         run_id: &str,
         moment: Moment,
-        script: &str,
+        script: String,
         timeout_seconds: u64,
     ) -> Result<ByteStream> {
         ensure_resource_supported(run_id, MIN_EXEC_VERSION, "command execution")?;
         let body = generated::types::ExecuteCommandRequest {
             moment,
-            script: script.to_string(),
+            script,
             timeout_seconds,
             // A server-side testing knob (routes output through the tenant
             // coordinator); snouty always wants the live session's output.
--- a/src/api.rs
+++ b/src/api.rs
@@ -924,15 +924,16 @@ const MIN_PROPERTIES_VERSION: u32 = 52;
 
 /// The tenant version that first served the run build logs resource. Runs
 /// created on older tenants 404 on `/runs/{run_id}/build_logs`.
-///
-/// Note the other nested run resources have different (or no) cutoffs: run
-/// properties arrived in v52, while logs and events are served for every
-/// version we can produce, so neither needs a guard.
 const MIN_BUILD_LOGS_VERSION: u32 = 54;
 
 /// The tenant version that first served the execute-command resource. Runs
 /// created on older tenants 404 on `/runs/{run_id}/execute_command`.
 const MIN_EXEC_VERSION: u32 = 60;
+
+// Every nested run resource with a version cutoff has a `MIN_*_VERSION` const
+// above, enforced by `ensure_resource_supported`. Logs and events have no
+// cutoff — they are served for every version we can produce — so neither
+// needs a guard. A new gated resource belongs in this group.
 
 /// Run IDs encode the tenant version that produced them as their second
 /// dash-delimited field — e.g. the `40` in
--- a/src/api.rs
+++ b/src/api.rs
@@ -326,25 +326,28 @@ impl AntithesisApi {
         }
     }
 
+    /// The endpoint takes each vtime as a decimal-seconds query parameter, so
+    /// the typed values are rendered with [`VTime`]'s exact `Display` — the
+    /// same text the value was parsed from, for anything snouty printed.
     pub async fn get_run_logs(
         &self,
         run_id: &str,
         input_hash: &str,
-        vtime: &str,
+        vtime: VTime,
         begin_input_hash: Option<&str>,
-        begin_vtime: Option<&str>,
+        begin_vtime: Option<VTime>,
     ) -> Result<ByteStream> {
         let mut request = self
             .client
             .get_run_logs()
             .run_id(run_id)
             .input_hash(input_hash)
-            .vtime(vtime);
+            .vtime(vtime.to_string());
         if let Some(v) = begin_input_hash {
             request = request.begin_input_hash(v);
         }
         if let Some(v) = begin_vtime {
-            request = request.begin_vtime(v);
+            request = request.begin_vtime(v.to_string());
         }
 
         match request.send().await {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -5,6 +5,17 @@ use crate::api::RunStatus;
 use crate::features::Feature;
 use crate::time::ReportDuration;
 
+/// clap value parser for a vtime the command passes through as text: the value
+/// must parse as a [`crate::vtime::VTime`], but the caller's exact spelling is
+/// what reaches the API. Commands that send a vtime as a typed field take
+/// `VTime` directly instead.
+fn validate_vtime(value: &str) -> Result<String, String> {
+    value
+        .parse::<crate::vtime::VTime>()
+        .map(|_| value.to_string())
+        .map_err(|err| err.to_string())
+}
+
 /// clap value parser for `--status` that keeps a friendly, enumerated error
 /// message (the generated `RunStatus::from_str` only says "invalid value").
 fn parse_run_status(value: &str) -> Result<RunStatus, String> {
@@ -512,7 +523,11 @@ pub struct DebugArgs {
     pub input_hash: Option<String>,
 
     /// Virtual time identifying the moment to debug
-    #[arg(long)]
+    // Validated as a vtime but kept as text: it is passed through verbatim as
+    // the `antithesis.debugging.vtime` parameter, and `--stdin` supplies the
+    // same parameter without going through this flag, so normalizing here
+    // would make the two paths disagree about what they send.
+    #[arg(long, value_parser = validate_vtime)]
     pub vtime: Option<String>,
 
     /// Debugging session description
@@ -621,12 +636,15 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
         input_hash: String,
 
         /// Virtual time of the moment to stream up to
+        // Typed, so a malformed vtime is rejected by clap instead of by the
+        // server. `allow_hyphen_values` is kept here (unlike `runs exec`),
+        // because this command has always accepted a hyphen-led vtime.
         #[arg(allow_hyphen_values = true)]
-        vtime: String,
+        vtime: crate::vtime::VTime,
 
         /// Start from this virtual time instead of the timeline's earliest log entry
         #[arg(long, allow_hyphen_values = true)]
-        begin_vtime: Option<String>,
+        begin_vtime: Option<crate::vtime::VTime>,
 
         /// Start from this input hash (optimization; must be paired with --begin-vtime)
         #[arg(long, allow_hyphen_values = true, requires = "begin_vtime")]
@@ -896,8 +914,11 @@ mod tests {
             panic!("expected `runs logs`");
         };
         assert_eq!(input_hash, "-123");
-        assert_eq!(vtime, "-2.0");
-        assert_eq!(begin_vtime.as_deref(), Some("-2.0"));
+        assert_eq!(vtime, "-2.0".parse::<crate::vtime::VTime>().unwrap());
+        assert_eq!(
+            begin_vtime,
+            Some("-2.0".parse::<crate::vtime::VTime>().unwrap())
+        );
         assert_eq!(begin_input_hash.as_deref(), Some("0"));
     }
 
@@ -913,7 +934,7 @@ mod tests {
             panic!("expected `runs logs`");
         };
         assert!(raw);
-        assert_eq!(vtime, "-2.0");
+        assert_eq!(vtime, "-2.0".parse::<crate::vtime::VTime>().unwrap());
 
         let cli = parse(&["snouty", "runs", "logs", "RUN", "-123", "-2.0"]);
         let Commands::Runs {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,7 +1,8 @@
 use chrono::{DateTime, Utc};
-use clap::{Args, Parser, Subcommand, ValueEnum};
+use clap::{Args, CommandFactory, Parser, Subcommand, ValueEnum};
 
 use crate::api::RunStatus;
+use crate::features::Feature;
 use crate::time::ReportDuration;
 
 /// clap value parser for `--status` that keeps a friendly, enumerated error
@@ -13,6 +14,75 @@ fn parse_run_status(value: &str) -> Result<RunStatus, String> {
              valid values: starting, in_progress, completed, cancelled, incomplete, unknown"
         )
     })
+}
+
+/// The global flags that decide which settings file and profile to read,
+/// read out of the raw arguments before the real parse.
+///
+/// A feature-gated subcommand has to be in the parser (or left out of it)
+/// *before* clap sees the arguments, so that `--help`, completions, and error
+/// messages all reflect the gate. But the gate is a setting, and which
+/// settings apply depends on `--settings`/`--profile` — which normally only
+/// the parse would tell us. This breaks the cycle by scanning for those two
+/// flags directly.
+///
+/// It is a scan rather than a second clap parser because both flags are
+/// `global = true`: they are accepted anywhere, including after a subcommand
+/// and its positionals, and no arrangement of a stand-in parser reproduces
+/// that without also having to know every subcommand's shape.
+#[derive(Default, Debug, PartialEq, Eq)]
+pub struct GlobalSettingsFlags {
+    pub settings: Option<std::path::PathBuf>,
+    pub profile: Option<String>,
+}
+
+impl GlobalSettingsFlags {
+    /// Read the two flags out of the real command line.
+    pub fn from_env() -> Self {
+        Self::from_args(std::env::args())
+    }
+
+    /// Read the two flags out of `args` (which includes the program name).
+    /// Never fails and never reports: a malformed command line is the real
+    /// parse's to complain about, and anything not understood here just
+    /// leaves the flag unset.
+    pub fn from_args(args: impl IntoIterator<Item = String>) -> Self {
+        let mut flags = Self::default();
+        let mut args = args.into_iter().skip(1);
+        while let Some(arg) = args.next() {
+            // A bare `--` ends the options; everything after it is a value,
+            // even if it is spelled like a flag.
+            if arg == "--" {
+                break;
+            }
+            // The last occurrence wins, matching clap for a non-repeating flag.
+            match take_flag_value(&arg, "--settings", &mut args) {
+                Some(value) => flags.settings = Some(value.into()),
+                None => {
+                    if let Some(value) = take_flag_value(&arg, "--profile", &mut args) {
+                        flags.profile = Some(value);
+                    }
+                }
+            }
+        }
+        flags
+    }
+}
+
+/// The value of `name` at `arg`, in either spelling: `--flag=value` carries it
+/// inline, `--flag value` takes the next argument.
+fn take_flag_value(
+    arg: &str,
+    name: &str,
+    rest: &mut impl Iterator<Item = String>,
+) -> Option<String> {
+    if let Some(value) = arg.strip_prefix(name).and_then(|v| v.strip_prefix('=')) {
+        return Some(value.to_string());
+    }
+    if arg == name {
+        return rest.next();
+    }
+    None
 }
 
 #[derive(Parser)]
@@ -112,8 +182,7 @@ Examples:
   snouty runs properties <run_id> --name <substring> --detail
   snouty runs build-logs <run_id>
   snouty runs logs <run_id> <hash> <vtime>
-  snouty runs events <run_id> -m <query>
-  snouty runs exec <run_id> <hash> <vtime> 'uname -a'"#,
+  snouty runs events <run_id> -m <query>"#,
         subcommand_required = false
     )]
     Runs {
@@ -639,8 +708,16 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
     },
 
     /// Execute a command in a run's live session
+    // Gated behind the `runs-exec` feature: hidden here, and revealed by
+    // `apply_feature_gates` only when the feature is enabled. Reaching it
+    // while disabled is refused in `main` (see `exec_is_gated_off`).
     #[command(
+        hide = true,
         long_about = r#"Execute a bash script in a run's live session, at a moment.
+
+This command is gated behind the `runs-exec` feature, because the Antithesis
+API it calls is unstable and unavailable on most tenants. Enable it with
+`features = ["runs-exec"]` in .snouty.toml, or SNOUTY_FEATURES=runs-exec.
 
 The run must have a live session (it is in progress). The script executes on
 a fresh branch of the multiverse, so it does not disturb the running test.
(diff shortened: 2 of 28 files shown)


src/api.rs
```diff
@@ -342,6 +342,32 @@ impl AntithesisApi {
...
+            timeout_seconds,
+            // A server-side testing knob (routes output through the tenant
+            // coordinator); snouty always wants the live session's output.
+            use_otis: false,
+        };
+        let request = self.client.execute_command().run_id(run_id).body(body);
+        match request.send().await {
+            Ok(response) => Ok(response.into_inner()),
+            Err(err) => Err(format_api_client_error(err).await),
+        }
```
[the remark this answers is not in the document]
carlsverre: The wire format was confirmed against the live API with this exact request path, in both forms. A curl probe with a bare-number vtime (`"vtime":16.299588482594118`) on the orbitinghail tenant returned 200 with correct output, and the finished snouty binary ran three end-to-end invocations (human mode, --json, stdin script) against a live run — all successful. The server accepts the number form.

Switching the request to the string form is also not the one-line change it looks like: `Moment.vtime` is generated as `VTime` via the build.rs conversion, and `VTime`'s number serialization is a deliberate, load-bearing choice (#192) that the `--json` output contract depends on. A string-serializing special case for this one request would need a parallel body type.

I added a doc comment on `execute_command` (be02dd5) that records the schema divergence and the live verification, so the next reader doesn't have to re-derive this.

src/api.rs
```diff
@@ -342,6 +342,32 @@ impl AntithesisApi {
...
+            moment,
+            script: script.to_string(),
+            timeout_seconds,
+            // A server-side testing knob (routes output through the tenant
+            // coordinator); snouty always wants the live session's output.
+            use_otis: false,
+        };
+        let request = self.client.execute_command().run_id(run_id).body(body);
+        match request.send().await {
+            Ok(response) => Ok(response.into_inner()),
```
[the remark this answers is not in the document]
carlsverre: Agreed — this is exactly the convention AGENTS.md pins. Fixed in be02dd5: `execute_command` now takes `script: String` and moves it into the request body.

src/runs.rs
```diff
@@ -1558,6 +1575,184 @@ async fn cmd_runs_logs(
...
+    debug!("executing command in run: {}", run_id);
+    let api = AntithesisApi::new_requiring_api_key(settings, verbose)?;
+    let moment = Moment {
+        input_hash: input_hash.to_string(),
+        vtime,
+    };
+    let stream = api
+        .execute_command(run_id, moment, &script, timeout)
+        .await?
+        .into_inner();
```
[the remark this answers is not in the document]
carlsverre: Agreed, and live probe data settles the open question in your analysis: the real endpoint's 404 body is `{"message":"Resource not found"}` — it never names the run — so exec would have shown a raw, inconsistent error where its siblings say `run not found: <id>`.

Fixed in be02dd5: the error now routes through `explain_run_scoped_error` like the other run-scoped streaming commands (the probe also keeps a real-run-but-404 case honest rather than guessing). The mock's exec 404 now reproduces the live body verbatim, and the spec asserts the translated message while forbidding the raw one, so the spec proves the translation instead of relying on a friendly mock body.

src/cli.rs
```diff
@@ -637,6 +638,47 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
...
+        /// Run ID
+        run_id: String,
+
+        /// Input hash of the moment to execute at (with VTIME, picks the timeline)
+        #[arg(allow_hyphen_values = true)]
+        input_hash: String,
+
+        /// Virtual time of the moment to execute at
+        #[arg(allow_hyphen_values = true)]
+        vtime: String,
```
carlsverre: can't we use our new VTime type?

src/cli.rs
```diff
@@ -637,6 +638,47 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
...
+    Exec {
+        /// Run ID
+        run_id: String,
+
+        /// Input hash of the moment to execute at (with VTIME, picks the timeline)
+        #[arg(allow_hyphen_values = true)]
+        input_hash: String,
+
+        /// Virtual time of the moment to execute at
+        #[arg(allow_hyphen_values = true)]
```
carlsverre: do we need allow hyphen values for vtime?

src/cli.rs
```diff
@@ -637,6 +638,47 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
...
+        vtime: String,
+
+        /// Bash script to execute; omit it (or pass `-`) to read the script
```
carlsverre: let's normalize to just omit for stdin rather than having two ways to trigger stdin reading

src/cli.rs
```diff
@@ -637,6 +638,47 @@ Output: `[vtime] [source] [stream] message`. A moment (HASH/VTIME) comes from
...
+        /// from stdin
+        script: Option<String>,
+
+        /// Maximum seconds the server waits for the script to exit before
+        /// reporting a timeout
+        #[arg(long, default_value_t = 30)]
```
carlsverre: does the server have a default value? min max value?

src/runs.rs
```diff
@@ -1558,6 +1584,153 @@ async fn cmd_runs_logs(
...
+        stream: String,
+        text: String,
+    },
+    Exited {
+        exit_code: Option<i64>,
+        /// Left untyped so a missing or malformed moment costs only the
+        /// trailer: typed here, it would fail the whole event and a real
+        /// `exited` would be misreported as a truncated stream.
+        #[serde(default)]
+        end_moment: Value,
```
carlsverre: did you observe this end moment being malformed or missing? why the extra caution? why not just `Option<Moment>`?

src/runs.rs
```diff
@@ -1558,6 +1584,153 @@ async fn cmd_runs_logs(
...
+            .suggestion("pass the script as an argument, or pipe it on stdin")
+    };
+    let script = match arg {
+        Some("-") | None => {
+            // Bare `runs exec RUN HASH VTIME` at an interactive terminal
+            // would silently wait for stdin; explain instead. An explicit `-`
+            // asked for stdin, so honor it either way.
+            if arg.is_none() && std::io::stdin().is_terminal() {
+                return Err(no_script());
+            }
```
carlsverre: does this mean you can't do `cat foo.sh | runs exec ...`?

src/cli.rs
carlsverre: or maybe we need this for a better interactive experience? reason about how to get the most idiomatic UX overall

src/runs.rs
```diff
@@ -1558,6 +1584,153 @@ async fn cmd_runs_logs(
...
+            let mut buf = String::new();
+            std::io::stdin()
+                .read_to_string(&mut buf)
```
carlsverre: you should add a sane limit here. big enough to accomodate any reasonable script, but small enough to not OOM the process if someone wires us up to devnull

carlsverre: 2MB is probably a fine starting point

src/runs.rs
```diff
@@ -1610,8 +1783,15 @@ fn format_log_entry(entry: &Value, raw: bool) -> String {
 /// vtime (see [`VTime`]). Returns the vtime, or `None` (entry untouched) when
 /// there's no parseable one.
 fn normalize_moment_vtime(entry: &mut Value) -> Option<VTime> {
-    let vtime = VTime::from_json(&entry["moment"]["vtime"])?;
-    entry["moment"]["vtime"] = json!(vtime);
+    normalize_vtime_field(entry, "moment")
+}
```
carlsverre: just inline uses of normalize_moment_vtime - rather than keeping around a one line wrapper

src/cli.rs
carlsverre: Yes — done in d0fea90. `vtime: crate::vtime::VTime` in the clap arg, so clap parses and rejects it and the hand-rolled parse in the dispatch arm is gone. The error is clap's standard form now: `invalid value 'not-a-number' for '<VTIME>': vtime is not a finite decimal number`. Spec updated to match.

Left `runs logs` alone — it passes vtime straight through as a query-param string and has tests pinning the current behavior, so converting it is a separate change.

src/cli.rs
carlsverre: No, and dropping it is an improvement — done in d0fea90 (kept on `input_hash`, which is routinely negative).

A vtime is seconds since the run began, so it is never negative; nothing in the spec or in any response I probed shows one. Keeping `allow_hyphen_values` had a real cost: it swallowed a misplaced flag as the vtime. Before: `snouty runs exec RUN HASH --timeout 45` → `invalid value '--timeout' for '<VTIME>'`. After: `the following required arguments were not provided: <VTIME>`.

src/cli.rs
carlsverre: Agreed — done in d0fea90. Omitting SCRIPT is now the only way to read stdin; `-` has no special meaning (it is just a script string). Help and spec updated.

src/runs.rs
carlsverre: No — `cat foo.sh | runs exec ...` works, and there is a spec case plus a unit test for it now.

The check only fires when stdin **is** a terminal, which means SCRIPT was omitted at a prompt with nothing piped in — the case that would otherwise hang with no indication it wants input. A pipe, a redirect (`< script.sh`), and a heredoc are all non-interactive, so they read normally. Verified against a live run: `echo 'uname -a' | snouty runs exec ...` and a heredoc both execute.

src/cli.rs
carlsverre: Server defaults: `timeout_seconds` defaults to **30**, `minimum: 0`, and there is **no maximum** in the spec.

So the CLI default of 30 mirrors the API's. In d0fea90 I added a floor of 1 (`value_parser = value_parser!(u64).range(1..)`) — a 0-second timeout is accepted by the server but can only ever time out, so it is always a mistake. No client-side ceiling, since the server documents none and guessing one could block a legitimately long command. All three facts are now in a comment next to the arg.

src/runs.rs
carlsverre: No, I never observed it — and the reasoning behind that caution was circular, so I have removed it. It is `Option<Moment>` now (d0fea90).

What happened: every live `exited` carried a full `end_moment`. I typed it first, and a test failed — but the failing input was `"end_moment":{}`, a fixture I had invented myself. I then wrote the comment to justify the untyped field. My own synthetic fixture drove the design, which is exactly the wrong reason.

The spec marks `end_moment` required on `exited`, so the mock now emits a real one everywhere and the invented malformed case is gone. `Option` remains only so a server that omits the field costs the trailer rather than the whole event.

src/runs.rs
carlsverre: Agreed — added in d0fea90.

carlsverre: 2 MiB it is: `MAX_SCRIPT_BYTES`, applied via `Read::take(limit + 1)` so an over-long script is refused rather than silently truncated into a script that means something different.

```
$ head -c 3000000 /dev/zero | tr '\\0' x | snouty runs exec RUN HASH VTIME
Error: the script on stdin is larger than the 2 MiB limit
Note: `runs exec` sends the whole script in one request body
```

Unit tests cover exactly-at-cap (accepted) and one byte over (refused).

src/runs.rs
carlsverre: Done in d0fea90 — `normalize_moment_vtime` is gone; both call sites use `normalize_vtime_field(entry, "moment" | "end_moment")` directly, and the doc comment moved onto the surviving function.

src/cli.rs
carlsverre: Reasoned through it and landed on: omit-only for stdin, and refuse to read from an interactive terminal. Both in d0fea90.

The argument for keeping `-` was the interactive case — with `-` gone and a bare invocation erroring at a TTY, there is no way to say "yes, I really do want to type a script now". But a heredoc covers that, and a heredoc is not a TTY:

```
snouty runs exec RUN HASH VTIME <<'SCRIPT'
ps aux | grep antithesis
SCRIPT
```

So the TTY refusal costs nothing real, while preventing the much more common accident — a bare `runs exec RUN HASH VTIME` sitting there looking hung. Typing a multi-line bash script blind into a CLI with no editor or readline was never a good experience anyway; the shell's heredoc is better at it than we would be.

That leaves one obvious way to do each thing: the argument for a one-liner, stdin (pipe, redirect, or heredoc) for anything longer. Verified all three against a live run.

carlsverre:
Added feature flags, and gated `runs exec` behind one (2164126).

The execute-command API is unstable and unavailable on most tenants, so the command ships in the next release turned off. A feature is enabled by id in the new `features` setting — a TOML array in a settings file, or a comma-separated `SNOUTY_FEATURES` — resolved through the same precedence chain as every other setting, so it can be set per profile:

```toml
# .snouty.toml
features = ["runs-exec"]
```

A gated-off command is absent, not merely hidden. It is missing from `--help` and from shell completions, and `snouty runs exec ...` (and `runs exec --help`) fails with clap's ordinary unrecognized-subcommand error, with nothing hinting that the command is real but disabled.

Two things were needed to get there. The gate has to be applied before clap parses, since it decides what the parser knows — but which settings apply depends on `--settings`/`--profile`, which normally only the parse reports. `GlobalSettingsFlags` breaks the cycle by scanning the raw arguments for those two flags; it is a scan rather than a second parser because both are `global = true` and so are valid after a subcommand and its positionals. And clap cannot remove a subcommand from a built `Command`, so a gated-off one is renamed to a parked name instead, which is what makes every path — invocation, `--help`, completions, error text — behave as if it were not there. Typing the parked name is refused too, since clap would otherwise accept it and the derive would quietly fall back to `runs` with no subcommand.

`snouty doctor` lists the enabled features, and omits the row when none are set. `specs/features.txt` covers the gate end to end: absence from help and completions, both settings spellings, per-profile resolution, malformed values, unknown ids carried rather than rejected, and the doctor row. `specs/runs_exec.txt` turns the feature on for the whole file.

Unknown ids are kept rather than rejected, so a settings file stays portable across snouty versions: an id from a newer build, or one whose feature has since graduated, does not break an older or newer binary reading the same file.

carlsverre:
Dropped the pre-parse scan: features now come from `SNOUTY_FEATURES` only (4fa5a3d).

You read the constraint right — the cycle was that the gate has to be known before the parse, while a setting can only be located after `--settings`/`--profile` are parsed. Making features env-only removes it outright, since an environment variable has no such dependency. The gate is now `hide = true` on the subcommand, `mut_subcommand(…, hide(false))` when the feature is on, and a refusal after the parse when it is off.

That deleted more than it added. `src/settings.rs` is reverted wholesale — the array-vs-string resolution path, the `lookup_setting` refactor, the `features` field, and the `assemble`-to-builder change all existed only to carry a TOML list, and none of it is needed now. `GlobalSettingsFlags` and its argument scanner are gone, `main` resolves settings inside `run` again as it did before, and the parked-name rename is gone with it. Net against the pre-feature commit: 406 insertions rather than the 615 the previous approach needed, and `settings.rs` is untouched.

Two consequences of hiding rather than removing, both verified and neither one something I can close within this design:

- `snouty runs exec --help` still prints the command's help when the feature is off. What it prints names the feature and how to enable it, and you have to already know the command to type that, so it reads as informative rather than leaky.
- Shell completions still offer `exec` when the feature is off. This one surprised me: `clap_complete` emits hidden subcommands into the candidate list for bash and zsh (fish omits them). Tab-completing it just leads to the refusal. `apply_feature_gates` is still applied to the generated script so it is right for generators that do respect `hide`.

I corrected both claims where I had asserted otherwise — the spec, the README, and a comment in `cmd_completions` had all said completions were gated, which was wrong. `specs/features.txt` now pins the real behaviour, including the two cases above, plus: refusal with clap's own error and its exit code 2, absence from `runs --help`, the comma-separated list, whitespace and empty entries dropped, unknown ids carried rather than rejected, an empty value meaning unset, a `features` key in a settings file having no effect, and the doctor row appearing only when something is set.

Verified against the orbitinghail tenant: refused with the variable unset, and a piped script executing normally with `SNOUTY_FEATURES=runs-exec`.

carlsverre:
It does, and that is better — the whole helper is gone (4cb5465).

clap-derive passes an attribute's value straight through to the builder method, so `hide` takes an expression, not just a literal, and it is evaluated when the command is built:

```rust
#[command(hide = !crate::features::is_enabled(Feature::RunsExec), long_about = …)]
Exec { … }
```

Since features come from the environment, that call needs nothing that only the parse could provide, so the decision happens inside `Cli::command()`. `apply_feature_gates` is deleted, `main` is back to a plain `Cli::parse()` with one added check, and `run` and `cmd_completions` no longer take a features parameter — `Cli::command()` already reflects the gate wherever it is called, including the completions generator.

What is left of the gate is two things: the `hide` expression, and `gated_command_error`, which turns an invocation of a gated-off command into clap's unrecognized-subcommand error. That second half is still needed because hiding does not make a subcommand uncallable.

One test went away rather than moving: the unit test asserting help visibility now depends on the process environment, which is racy under a parallel test runner. `specs/features.txt` already covers both directions in subprocesses with the variable set, which is the right place for it. The remaining unit test on `gated_command_error` stays pure — it takes the enabled features as a parameter.

Net against the pre-feature commit is now 368 insertions across 8 files, with `settings.rs` untouched and `main.rs` up by 15 lines. Verified live on the orbitinghail tenant: hidden from `runs --help` and refused with exit 2 when unset, listed and executing normally with `SNOUTY_FEATURES=runs-exec`.

carlsverre:
Every command-line vtime now goes through the `VTime` type (7516ab0, 92d33d6). There were three besides `runs exec`.

**`runs logs`** — the positional `VTIME` and `--begin-vtime` are typed `VTime` fields. They reach the API as decimal-seconds query parameters, rendered with `VTime`'s exact `Display`, and I confirmed with `--verbose` that a full-precision vtime goes out byte-identical (`vtime=18.030696514528245`), plus a live `runs logs` call still streaming. `allow_hyphen_values` stays there, unlike on `runs exec`: this command has always accepted a hyphen-led vtime and a test pins it, so narrowing that is a separate decision.

**`debug --vtime`** — typed `Option<VTime>`. As you said, the server is fine with `123` becoming `123.0`.

Typing the flag alone would have left the command inconsistent with itself, though, because a debug moment arrives by three routes and they spelled the same moment differently. `Moment.from` on stdin already normalized through `VTime` (from #192), the typed flag did not, and raw JSON on stdin passed its text through untouched. So the rule now lives in one place — `Params::normalize_debug_vtime`, applied in `debug_params` after the routes merge — and `moment.rs` drops its own copy of it, going back to plain "numbers become strings".

All three routes now agree, verified by hand:

```
--vtime 402                                    -> "antithesis.debugging.vtime": "402.0"
Moment.from({ …, vtime: 402 })                 -> "antithesis.debugging.vtime": "402.0"
{"antithesis.debugging.vtime": "402"}          -> "antithesis.debugging.vtime": "402.0"
--vtime 329.8037810830865                      -> "antithesis.debugging.vtime": "329.8037810830865"
```

A malformed vtime is now rejected before any API call on every route, where the raw-JSON one previously reached the server:

```
$ snouty debug --run-id r-1 --input-hash 42 --vtime abc
error: invalid value 'abc' for '--vtime <VTIME>': vtime is not a finite decimal number

$ echo '{"antithesis.debugging.vtime": "not-a-number", …}' | snouty debug --stdin
Error: invalid arguments: vtime is not a finite decimal number
Note: antithesis.debugging.vtime = "not-a-number"
```

Three assertions moved to the normalized text: two in `specs/debug.txt` and one in `tests/cli_general.rs`. New coverage for the malformed cases in `specs/debug.txt` and `specs/runs.txt`, and unit tests on `normalize_debug_vtime` for the spelling agreement, the rejections, and an absent vtime still being the schema's error to report rather than this function's.

carlsverre:
@claude review

src/runs.rs
```diff
@@ -1558,6 +1575,178 @@ async fn cmd_runs_logs(
...
+    match arg {
+        Some(script) => {
+            if script.trim().is_empty() {
+                return Err(no_script_error());
+            }
+            Ok(script.to_string())
+        }
+        None => read_script_from_stdin(std::io::stdin(), std::io::stdin().is_terminal()),
+    }
+}
```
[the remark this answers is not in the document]
carlsverre: Agreed — fixed in e2375fa. `cmd_runs_exec` and `resolve_exec_script` now take `Option<String>` and the owned value moves straight through from the `cmd_runs` match arm, so the `.to_string()` is gone. Same convention as the `execute_command` fix in be02dd5.

src/runs.rs
```diff
@@ -1558,6 +1575,178 @@ async fn cmd_runs_logs(
...
+            "the script on stdin is larger than the {} MiB limit",
+            MAX_SCRIPT_BYTES / (1024 * 1024)
+        ))
+        .note("`runs exec` sends the whole script in one request body"));
+    }
+    if buf.trim().is_empty() {
+        return Err(no_script_error());
+    }
+    Ok(buf)
+}
```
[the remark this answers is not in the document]
carlsverre: Good catch, and it reproduces exactly as described — fixed in e2375fa.

I had hand-checked the cap when I added it, but only with all-ASCII input (`tr '\0' 'x'`), which is precisely the case that works. Both non-ASCII forms fail:

```
$ head -c 3000000 /dev/urandom | snouty runs exec …
Error:
   0: failed to read the script from stdin
   1: stream did not contain valid UTF-8

$ python3 -c "…'é'.encode()*1500000" | snouty runs exec …
Error:
   0: failed to read the script from stdin
   1: stream did not contain valid UTF-8
```

`MAX_SCRIPT_BYTES + 1` is odd, so a buffer of 2-byte characters splits one at the boundary every time — not just half the time.

The fix reads into a `Vec<u8>` with `read_to_end`, checks the size, and converts with `String::from_utf8` only after that. Both cases now report the limit:

```
Error: the script on stdin is larger than the 2 MiB limit
Note: `runs exec` sends the whole script in one request body
```

One thing beyond the reported bug: a *sub*-cap non-UTF-8 script previously surfaced the raw io error too. It now gets its own message, `the script on stdin is not valid UTF-8`, rather than a generic read failure.

Two regression tests cover what the old ones missed — the ASCII-only fixtures were why this slipped through. One drives an over-cap buffer of multi-byte characters and one an over-cap binary buffer, both asserting the size message; a third asserts the sub-cap invalid-UTF-8 message.

src/api.rs
```diff
@@ -326,25 +326,28 @@ impl AntithesisApi {
...
-        begin_vtime: Option<&str>,
+        begin_vtime: Option<VTime>,
     ) -> Result<ByteStream> {
         let mut request = self
             .client
             .get_run_logs()
             .run_id(run_id)
             .input_hash(input_hash)
-            .vtime(vtime);
+            .vtime(vtime.to_string());
```
carlsverre: why does request.vtime and begin_vtime not take an instance of VTime?

src/api.rs
```diff
@@ -353,6 +356,37 @@ impl AntithesisApi {
...
+        script: String,
+        timeout_seconds: u64,
+    ) -> Result<ByteStream> {
+        let body = generated::types::ExecuteCommandRequest {
+            moment,
+            script,
+            timeout_seconds,
+            // A server-side testing knob (routes output through the tenant
+            // coordinator); snouty always wants the live session's output.
+            use_otis: false,
```
carlsverre: can we leave this out of the request entirely rather than setting it? i'd rather rely on the server to default it since the server is likely to remove this field

src/api.rs
```diff
@@ -892,12 +926,19 @@ const MIN_PROPERTIES_VERSION: u32 = 52;
...
+// Every nested run resource with a version cutoff has a `MIN_*_VERSION` const
+// above, enforced by `ensure_resource_supported`. Logs and events have no
+// cutoff — they are served for every version we can produce — so neither
+// needs a guard.
+//
+// Execute-command has no cutoff either, on purpose. The multiverse debugger
+// has not shipped, so the release that first serves the endpoint is not
+// settled, and a pinned floor here would reject runs the server can in fact
+// execute against. The feature is unstable, so the server's own answer is the
+// authority. Add a const here once the release is known.
```
carlsverre: delete this

src/cli.rs
```diff
@@ -704,6 +769,38 @@ impl Default for RunsListArgs {
...
+    // (`snouty runs`) rather than the bare subcommand name.
+    cli.build();
+    Some(
+        cli.find_subcommand_mut("runs")
+            .expect("the parent of a gated subcommand exists")
+            .error(
+                clap::error::ErrorKind::InvalidSubcommand,
+                "unrecognized subcommand 'exec'",
+            ),
+    )
```
carlsverre: this is code smell. we don't need the error to be idental to what clap would say if the command didn't exist. that's way too overkill. Just emit a nice, clear error that says the command is unstable and must be enabled via `command --help`. if someone knows to run the command they know it exists. giving them a 404 is just annoying at that point

src/runs.rs
```diff
@@ -1558,6 +1566,184 @@ async fn cmd_runs_logs(
...
+/// Hand-written rather than the generated `ExecuteCommandStreamEvent`: that
+/// type is an untagged enum whose variants progenitor names `Variant0/1/2`,
+/// which reads far worse at the match sites than the three named variants
+/// here.
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+enum ExecEvent {
+    Output {
+        stream: String,
+        text: String,
```
carlsverre, line 1604: we should keep the output event's moment. we will render it in the json output version of exec

src/runs.rs
```diff
@@ -1558,6 +1566,184 @@ async fn cmd_runs_logs(
...
+/// output event's `moment`) are ignored; the `exited` event's `end_moment` is
+/// kept for the end-moment trailer.
+///
+/// Hand-written rather than the generated `ExecuteCommandStreamEvent`: that
+/// type is an untagged enum whose variants progenitor names `Variant0/1/2`,
+/// which reads far worse at the match sites than the three named variants
+/// here.
+#[derive(Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+enum ExecEvent {
```
carlsverre, line 1601: I think we should make this type more flexible. ideally collect unknown frames into a catchall UnknownEvent(serde_json::Value) and unknown params into a map.

this is due to the exec command still being a WIP and may evolve a bunch as we iterate on it. just leave a comment to lock this down to an explicit type when the command is stabilized

src/api.rs
carlsverre: Done in ec1239f — snouty now sends no `use_otis` at all.

Setting it to `false` was unavoidable at the struct level: the generated `ExecuteCommandRequest` has no `skip_serializing_if`, so any value serializes. So `build.rs` now drops the field from the schema before generation (`drop_use_otis`), next to the other spec transforms. The generated type has no such field, and the wire body cannot carry one.

The transform asserts the pointer exists, so a spec refresh that retires the field fails the build with a message telling the next person to delete the transform.

src/api.rs
carlsverre: They do now — ec1239f. `.vtime(vtime)` and `.begin_vtime(v)`, no `to_string()` at the call site.

The setters are generic over `TryInto<String>`, so the fix was an `impl From<VTime> for String` that goes through `Display`.

Worth recording why the conversion is not `Serialize`: query parameters serialize through serde, and `VTime`'s `Serialize` writes a JSON *number*. Routing the wire text through `serialize_f64` would make a second, unpinned rendering of a precision-critical value. `Display` is the one snouty pins everywhere else, so the conversion uses it, and a test asserts the round trip over every full-precision vtime in the fixtures.

src/api.rs
carlsverre: Deleted in ec1239f.

src/cli.rs
carlsverre: Agreed, and you are right about the reason — pretending the command is absent only wastes the time of someone who already knows it exists. Replaced in ec1239f:

```
$ snouty runs exec RUN HASH VTIME 'uname -a'
Error: `snouty runs exec` is an unstable feature and is not enabled

Note: enable it by setting SNOUTY_UNSTABLE_FEATURES=runs-exec
Note: an unstable feature can change or go away in any release
Suggestion: run `snouty runs exec --help` for what it does
```

This deleted the whole clap-mimicking apparatus: the `Cli::command()` rebuild, the `build()` call that existed only to fix the usage line, and the `find_subcommand_mut` lookup. `gated_command_error` now returns a `Report` built with `user_error`, so it travels the same error path as every other snouty failure and exits 1 rather than clap's 2.

The gate still hides the command from `--help` and from listings. Only the refusal changed.

Two spec cases moved with it. `specs/features.txt` step 1 now asserts the new message, and step 3 changed meaning: `runs exec --help` prints the help rather than being refused, which is what the error's suggestion points at.

src/runs.rs
carlsverre: Done in ec1239f, both parts.

An outer `ExecFrame` is untagged, so a frame this build does not know is kept whole:

```rust
enum ExecFrame {
    Known(ExecEvent),
    Unknown(Value),
}
```

Each known variant gains `#[serde(flatten)] extra: Map<String, Value>`, so a field the server adds is kept instead of dropped. The `output` variant also keeps its `moment` now, per your other comment.

The doc comment says why the type is loose and when to tighten it: the command is a work in progress behind an unstable feature, so the stream may still grow frames and fields, and this should become a closed type once the command stabilizes.

On your follow-up about `--json`: that already held, and now a spec pins it. `--json` prints the whole normalized line before any typing, so an unknown frame and an unknown field both reach stdout unchanged. The mock grows an `unknown-frames` script that emits a `heartbeat` frame and an `output` frame carrying `truncated`. The spec asserts that `--json` passes both through, and that in human mode the unknown frame goes to stderr while the known frame still renders its text — an unrecognized frame is not command output, so it must not land on stdout.

The retained fields are not read yet, so they carry an `#[allow(dead_code)]` with a comment stating that they exist to survive into the rendered JSON form rather than being dropped here.

src/runs.rs
carlsverre: Kept in ec1239f. The `output` variant now has `moment: Option<Moment>`, and a unit test asserts it parses with full precision rather than being discarded.

It is `Option` for the same reason as `end_moment`: while the command is unstable, a missing field should cost the moment, not the whole frame.

Nothing reads it yet, so it sits under the `#[allow(dead_code)]` described in the sibling thread, with a comment saying it is there for the rendered JSON form. The raw `--json` output already carries the moment today, because that path prints the line before any typing.

src/runs.rs
```diff
@@ -1687,21 +1720,23 @@ async fn cmd_runs_exec(
                 if json {
                     outln!("{entry}")?;
                 }
-                match ExecEvent::deserialize(&entry).ok() {
-                    Some(ExecEvent::Output { stream, text }) if !json => {
+                match ExecFrame::deserialize(&entry) {
```
carlsverre: Putting `!json` guards on every case just to extract a terminal event is a code smell and makes it easy to break in the future.

The code should be simpler and more straightforward, and clearly delineated based on whether you are in JSON output mode or not. Figure out a nicer way to thread this code so it's easier to maintain.

src/runs.rs
```diff
@@ -1687,21 +1720,23 @@ async fn cmd_runs_exec(
...
+                    )) => {
                         terminal = Some(event);
                     }
-                    // An unrecognized event is not command output; keep it
-                    // off stdout (--json already passed the line through).
-                    None if !json => eprintln!("{entry}"),
+                    // A frame this build doesn't know is not command output;
+                    // keep it off stdout (--json already passed it through).
+                    Ok(ExecFrame::Unknown(_)) | Err(_) if !json => eprintln!("{entry}"),
                     _ => {}
```
carlsverre: presumably if we fail to decode we should error...

src/runs.rs
carlsverre: Agreed — the guards had spread across the arms until the mode and the frame kind were tangled together. Restructured in 4eee207 so each is decided once:

```rust
let frame = ExecFrame::deserialize(&entry).map_err(…)?;

if json {
    outln!("{entry}")?;
} else {
    render_exec_frame(&frame)?;
}

// Either mode: the terminal event decides the exit status.
if let ExecFrame::Known(event @ (ExecEvent::Exited { .. } | ExecEvent::TimedOut { .. })) = frame {
    terminal = Some(event);
}
```

One `if json`, for rendering only. Capturing the terminal event no longer sits inside the match over output shapes, so the thing that decides the exit status is stated once and cannot be missed by a future arm. No arm carries a guard.

Human rendering moved into `render_exec_frame`, which matches on the frame kind and knows nothing about the mode. That is where the "only the script's own output reaches stdout" rule now lives, with the reason next to it.

The split also removed the `#[allow(dead_code)]` on `ExecFrame`: `Unknown` is read by the renderer now, so only the retained `moment`/`extra` fields still need it.

src/runs.rs
carlsverre: Right — it errors now (4eee207):

```rust
let frame = ExecFrame::deserialize(&entry).map_err(|err| {
    eyre!("could not decode a line of the command's output: {err}")
        .note("every JSON object decodes, so this means the decoder itself failed")
})?;
```

Previously a decode failure shared an arm with an unknown frame and was printed to stderr, which would have hidden a real fault behind output that looks like ordinary passthrough.

Worth stating plainly though: this arm should never run. `ExecFrame` is untagged with an `Unknown(serde_json::Value)` fallback, and `Value` accepts any JSON, so no shape the server sends can fail — and `serde_json` is built without `arbitrary_precision`, which is the one feature that would break untagged decoding into `Value`. A new test drives six hostile shapes (empty object, wrong types, a null tag, a numeric tag, nested junk) and asserts each decodes, so the reason the error path is unreachable is pinned rather than assumed. Failing loudly is still the right handling for a case that cannot happen.

PR #212, opened 2026-08-08
Deprecate username/password auth, move login prompts to inquire, add PTY tests

carlsverre:
This PR starts the deprecation of username/password authentication and reworks the interactive login prompts.

`snouty launch` and `snouty debug` print a warning when they proceed with username/password credentials. The warning tells the user to run `snouty login` to switch to another authentication method, and `snouty login` shows which methods the tenant offers. Commands that reject username/password authentication add the same suggestion to their error, `snouty doctor` now calls the method deprecated instead of legacy, and the login menu labels the option "(deprecated)".

The auth constructors get names that match their intent. `AntithesisApi::new_requiring_api_key` did not require an API key: it accepted API key, OAuth, and OIDC credentials, and only rejected username/password. That strict behavior is now `AntithesisApi::new`, the default, so new call sites get the safe variant. The old permissive `new` is now `new_for_launch`, named for the launch webhooks — the only endpoints that still accept username/password. `allow_basic` becomes `allow_password`.

The login prompts move from dialoguer to inquire, which deletes the hand-rolled terminal code and fixes four problems:

- The credential menu highlighted nothing by default (dialoguer treats an unset default as `usize::MAX`). The menu now always highlights the previously used method, or the first option.
- Single sign-on leads the menu and is the first-login default when the tenant offers it; the deprecated username/password option comes last.
- Secret prompts echo `*` per character (dialoguer's `Password` echoes nothing, which makes a pasted API key look like a hang). inquire renders long masked input as a scrolling one-row viewport.
- A long default value (e.g. a registry URL) wrapped the prompt line, which dialoguer re-renders duplicated on submit. inquire's frame renderer handles wrapped rows.

dialoguer is removed; console stays for non-interactive styling, width probing, and table truncation, which crossterm does not cover.

New `tests/login_pty.rs` integration tests drive `snouty login` end-to-end through a real pseudo-terminal via expectrl: menu default selection, star masking for a pasted key, a key wider than the terminal (including backspace edits), and the single-entry password flow. The password prompt asks once, with no confirmation round: an Antithesis password is a long generated string that is pasted like an API key, not typed twice. The spec harness runs commands without a TTY, so interactive coverage lives here; the non-interactive login paths stay covered by `specs/login.txt`. Every send is gated on an expect that proves its prompt is rendered, which makes the dialogue deterministic. On an expect timeout, the harness drains the unmatched output (the pattern recommended in zhiburt/expectrl#75), strips ANSI escapes with strip-ansi-escapes, and panics with it, so a CI failure shows what the child had on screen.

Specs for `launch`, `debug`, `runs`, and `doctor` assert the new warning and error text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 # Unreleased
 
-- Deprecate username/password authentication: `snouty launch` and `snouty debug` print a warning that steers the user to `snouty login` (single sign-on or an API key)
-- Improve the `snouty login` prompts: single sign-on leads the credential menu and the first option is preselected, secret prompts echo `*` per character, and a long current value no longer breaks the prompt rendering
-- Add `snouty login`: interactive sign-in with browser-based OAuth or an API key, multiple profiles, and credential storage in the macOS keychain or a credentials file on Linux ([#157](https://github.com/antithesishq/snouty/pull/157), [#173](https://github.com/antithesishq/snouty/pull/173), [#178](https://github.com/antithesishq/snouty/pull/178))
+- Deprecate username/password authentication: `snouty launch` and `snouty debug` print a warning that steers the user to `snouty login` ([#212](https://github.com/antithesishq/snouty/pull/212))
+- Add `snouty login`: interactive sign-in with browser-based OAuth or an API key, multiple profiles, and credential storage in the macOS keychain or a credentials file on Linux ([#157](https://github.com/antithesishq/snouty/pull/157), [#173](https://github.com/antithesishq/snouty/pull/173), [#178](https://github.com/antithesishq/snouty/pull/178), [#212](https://github.com/antithesishq/snouty/pull/212))
 - Add GitHub Actions OIDC authentication ([#157](https://github.com/antithesishq/snouty/pull/157))
 - Add support for the `docker compose` plugin in addition to standalone `docker-compose` ([#172](https://github.com/antithesishq/snouty/pull/172))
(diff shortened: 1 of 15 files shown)


carlsverre:
Both review nits are addressed in 34154ab: the deprecation sentence now lives in one shared constant (`PASSWORD_DEPRECATION_SUGGESTION` in auth.rs, referenced by the rejection error and the doctor note), and the stale "legacy" comment in doctor.rs now uses the deprecation wording.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Deprecate username/password authentication: `snouty launch` and `snouty debug` print a warning that steers the user to `snouty login` (single sign-on or an API key)
+- Improve the `snouty login` prompts: single sign-on leads the credential menu and the first option is preselected, secret prompts echo `*` per character, and a long current value no longer breaks the prompt rendering
 - Add `snouty login`: interactive sign-in with browser-based OAuth or an API key, multiple profiles, and credential storage in the macOS keychain or a credentials file on Linux ([#157](https://github.com/antithesishq/snouty/pull/157), [#173](https://github.com/antithesishq/snouty/pull/173), [#178](https://github.com/antithesishq/snouty/pull/178))
```
carlsverre: merge these two changelog entries into one. snouty login has not made it into a full release yet, so these should just read as a single feature. also implementation details like "single sign-on leads the credential menu and the first option is preselected, secret prompts echo `*` per character, and a long current value no longer breaks the prompt rendering" honeslty don't make sense in a changelog. the changelog should be focused on new and updated behaviors at a high level, not enumerating every small implementation detail.

carlsverre: Fixed in 45f07c4. The prompt-details entry is gone; its PR number joins the single `snouty login` feature entry, and the deprecation entry keeps only the high-level behavior.

PR #214, opened 2026-08-10
Stream validate's container logs in real order, and stop masking compose failures

carlsverre:
## Why

`snouty validate` started containers with `up --detach`, then followed `compose logs`. That replays each container's buffered history as its own block, so the startup window — the part worth reading — arrived grouped by container rather than in the order things actually happened.

Measured on a three-service project logging concurrently every 50ms:

| | max displacement | longest single-container run | first 12 lines |
|---|---|---|---|
| before | **120** of 180 | **59** of 60 | `C1 C2 C3 C4 C5 C6 C7 A1 A2 A3 A4 A5` |
| after | **0** | **1** | `A1 C1 B1 A2 C2 B2 A3 C3 B3 A4 C4 B4` |

This came from a user report that validate wasn't honouring `depends_on`. It was — their graph never connected the two services they were comparing. But the log replay made a correct startup look broken, which is a bug worth removing on its own.

(Adjacent-inversion count is a trap here: the old output scored only 5.6% while being maximally grouped, because three blocks means about two seams. Displacement and longest-run are the meaningful metrics.)

## What changed

**Attached `compose up`.** Only an attached `up` interleaves live output. stdio is still inherited so compose keeps its per-service colors, and stdin stays null so it can't decide it's interactive and block on a prompt — the failure that got the original PTY-based attached `up` replaced in 990e484. Deliberately no `--abort-on-container-exit`: a one-shot setup container exiting 0 is a legitimate pattern, and the `stranded` fixture covers it.

**One timeout budget.** An attached `up` never returns, so `--timeout` now covers startup *and* the setup-complete wait, with the default raised 60 → 120 to match the old combined ceiling (it was previously spent twice).

**Discovery keeps its fast-fail.** It only needs a container to *exist* — `cp` reads templates from a container that never started — and compose creates the whole project up front, so a short existence poll restores the ordering `up --detach` used to provide. Verified: a misnamed test command still fails in ~2s, not after the full timeout.

**The hermetic check no longer masks unrelated failures.** It reported *any* failed render as \"depends on your shell environment\". Since it cleared `PATH` and compose shells out to \`docker\`, every compose file on a standalone compose install failed that way, sending users hunting through a file that was never the problem. `PATH` is no longer scrubbed — it's machinery for running compose, not a value users interpolate, and Antithesis has a `PATH` too — and only an interpolation failure now counts as divergence.

**Compose floor at v2.24.7.** snouty has always needed `--no-path-resolution`; without it, relative build contexts and bind-mount sources are rewritten to absolute paths on the developer's machine and baked into the config image. The flag arrived in [v2.18.0](https://github.com/docker/compose/pull/10557) but was ignored for files using `include:` until [v2.24.7](https://github.com/docker/compose/issues/11508), so anything lower can still silently ship local paths. Previously this surfaced as a raw `unknown flag` error from `snouty launch`.

## Tests

Engine spec timeouts doubled to preserve their effective headroom under the single budget. Launch fixtures now build `--platform linux/amd64` explicitly rather than inheriting the build host's arch — the negative arm64 case is untouched.

**611/611 pass, 0 skipped.** Worth noting the baseline: 11 of these were failing on my machine before this branch, all from a Docker Desktop too old to run them. Everything here was verified against compose 5.3.1 / Docker 29.6.2.


carlsverre:
@claude review

src/validate.rs
```diff
@@ -256,61 +259,46 @@ async fn validate_compose(
...
+        _ = tokio::signal::ctrl_c() => Err(eyre!("interrupted")),
+    };
+    let scripts = match ready.and_then(|()| discover_scripts(rt, &compose, overlay, temp_dir)) {
+        Ok(scripts) => scripts,
+        Err(e) => {
             up_child.kill_group().await.ok();
-            bail!("interrupted");
+            return Err(e);
         }
     };
```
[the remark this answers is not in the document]
carlsverre: Fixed in 7c44288. `compose ps` now reports a strict `running` flag on `ComposeContainer`, and `discover_scripts` passes that instead of `!stopped`. A `created` container now takes the `cp` path, which works in every container state.

src/compose.rs
```diff
@@ -170,17 +172,53 @@ impl ComposeCli {
...
+/// Antithesis, where they mean nothing.
+///
+/// The bar is v2.24.7 rather than v2.18.0 because the flag was ignored for
+/// files pulled in with `include:` until then (docker/compose#11508). Since
+/// `include:` resolves each file's relative paths against that file's own
+/// directory, a project using it on v2.20 (where `include:` landed) through
+/// v2.24.6 would pass a looser check and still ship absolute local paths —
+/// silently, and only visible once the run is on the platform. A version check
+/// that lets a broken config through is worse than no check.
+const MIN_COMPOSE_VERSION: (u64, u64, u64) = (2, 24, 7);
```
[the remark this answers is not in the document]
carlsverre: Fixed in d8abd17.

src/validate.rs
```diff
@@ -321,6 +309,48 @@ async fn validate_compose(
...
+            if services.iter().all(|s| present.contains(s.as_str())) {
+                return Ok(());
+            }
+        }
+        if tokio::time::Instant::now() >= deadline {
+            bail!("timed out waiting for compose to create containers");
+        }
+        sleep(Duration::from_millis(100)).await;
+    }
+}
```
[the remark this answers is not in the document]
carlsverre: Fixed in 19fa5cf. `validate` now checks `compose ps -a` before `up` and refuses to start when the project already has containers. The error prints the exact `compose down` command to run. This closes the window where leftover containers satisfy the readiness poll.

src/compose.rs
```diff
@@ -894,6 +956,13 @@ pub struct ComposeContainer {
     /// `created`, `restarting`, `paused`, and missing State are treated as
     /// not-stopped to avoid false positives on healthy setups still settling.
     pub stopped: bool,
+    /// Whether the container's entrypoint is currently running, i.e. `exec`
+    /// can work in it. Strictly the `running` state: a `created` container
+    /// is neither stopped nor running (compose creates containers up front
+    /// and starts them as `depends_on` conditions clear), and `exec` fails
+    /// in it, while `cp` works — so discovery must not equate "not stopped"
+    /// with "exec-able".
+    pub running: bool,
```
carlsverre: these two bools should be combined into a single state enum

src/validate.rs
```diff
@@ -256,61 +285,46 @@ async fn validate_compose(
...
+        status = up_child.wait() => Err(compose_exited_early(status)),
+        _ = tokio::signal::ctrl_c() => Err(eyre!("interrupted")),
+    };
+    let scripts = match ready.and_then(|()| discover_scripts(rt, &compose, overlay, temp_dir)) {
+        Ok(scripts) => scripts,
+        Err(e) => {
             up_child.kill_group().await.ok();
-            bail!("interrupted");
+            return Err(e);
         }
```
carlsverre: do test command discovery after setup complete. fix the test to use a shorter timeout or make setup complete fire very quickly.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+- `snouty validate` no longer reports an unrelated `docker-compose config` failure as "depends on your shell environment". Only a genuine interpolation failure counts as an environment dependency; anything else surfaces as itself. `PATH` is also no longer scrubbed during that check — compose shells out to `docker`, so scrubbing it made every compose file fail the check on a standalone compose install
+- `snouty validate` fails fast when the compose project already has containers — leftovers from a `--keep-running` session or a crashed run — and prints the exact `compose down` command to run first
+- `snouty validate` streams container logs in the order things actually happened. It previously started containers detached and then replayed their logs, which printed each container's history as its own block; the startup sequence now interleaves correctly. `--timeout` is now a single budget covering both container startup and the wait for setup-complete (previously each got its own), and its default rises from 60 to 120 seconds to match the old combined ceiling
```
carlsverre: reduce these changelog entries to something simpler and more abstract

carlsverre: also include the PR link (214)

src/compose.rs
carlsverre: Fixed in 94b96db: `ComposeContainer` now carries a three-variant `ContainerState` enum (Running / Stopped / Transient).

src/validate.rs
carlsverre: Fixed in e39f724. Discovery now runs after setup-complete, and `wait_for_service_containers` is deleted. No test changes were needed: every engine-spec fixture emits setup-complete immediately, and the validate specs pass in ~13s.

CHANGELOG.md
carlsverre: Fixed in ff4683a: reduced to two one-line entries, each linking #214.

PR #218, opened 2026-08-13
validate: read SDK output from the start, and explain a rewrite

### LINKED ISSUE #216: validate: sdk.jsonl is still shared per-service, so replicas and multi-process containers can truncate setup-complete
Follow-up to #64. That fix gives each compose *service* its own subdir of the bind mount, which covers the cross-service case. Two cases remain, because the subdir is keyed by service name and the SDK truncates the file on open.

### 1. Replicas of one service share a file

The generated override keys the host path by service name:

```yaml
services:
  twin:
    volumes:
    - /tmp/.tmp2RRKqa/antithesis/twin:/tmp/antithesis:z
    environment:
      ANTITHESIS_OUTPUT_DIR: /tmp/antithesis
      ANTITHESIS_SDK_LOCAL_OUTPUT: /tmp/antithesis/sdk.jsonl
```

Repro:

```yaml
services:
  twin:
    image: registry:2
    deploy:
      replicas: 2
    entrypoint: ["sh","-c","echo \"line from $$(hostname)\" >> \"$$ANTITHESIS_OUTPUT_DIR/sdk.jsonl\"; sleep 300"]
```

One directory is created for the service, and both containers write into it:

```
$ find /tmp/.tmp2RRKqa/antithesis -mindepth 1
/tmp/.tmp2RRKqa/antithesis/twin
/tmp/.tmp2RRKqa/antithesis/twin/sdk.jsonl

$ cat /tmp/.tmp2RRKqa/antithesis/twin/sdk.jsonl
line from 2a3ca87eb422
line from 8b814f773827
```

### 2. Multiple SDK-linked processes in one container share a file

A per-service subdir does not separate processes within a container. An entrypoint plus test commands in the same container all resolve to the same `ANTITHESIS_SDK_LOCAL_OUTPUT`.

### Mechanism

Same as #64. `antithesis-sdk-go` `internal/emit.go` (verified on released v0.7.0):

```go
file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0644)
if err != nil {
    ...
} else if err = file.Truncate(0); err != nil {
```

There is no `O_APPEND`, and `Truncate(0)` runs during package init. Each SDK-linked process empties `sdk.jsonl` when it starts and writes from offset 0, and concurrent writers hold independent file offsets. An already-written setup-complete event is lost when the next process initializes, so `validate` can report `timed out waiting for setup-complete event` for a system that did emit it.

Only the local-file path is affected. `init_in_antithesis()` takes precedence when available, so real runs use the voidstar handler and no file is involved.

Observed with snouty 0.7.0-rc.1 (864350a), docker fallback (`podman not found`), Linux.

carlsverre:
`snouty validate` could time out while the setup-complete event was sitting on disk.

`watch_for_setup_complete` tailed each `.jsonl` file from a stored offset, and reset that offset only when the file became shorter than it. The SDK's local-file handler opens its output without `O_APPEND` and truncates it during initialization, so a second SDK-linked process writes from offset 0 into bytes snouty had already consumed. When the rewrite left the file no shorter, the offset stood still and snouty never re-read the region that now held the event.

The fix re-reads each file from the start whenever its length or mtime moved. `contains_setup_complete` no longer seeks back over a partial line, so it takes the content directly — an unfinished line fails to parse and is read again on the next poll.

`watch_for_setup_complete` now returns a `SetupOutcome` enum, so the call site reads as `Found`, `TimedOut`, or `Err` rather than `Ok(true)` and `Ok(false)`.

The poll interval drops from 100ms to 25ms, which bounds how long a written event can sit unread before another process truncates it away. Native filesystem notifications would shorten that window by roughly a further 20ms, but `notify`'s own documentation records that no backend is a fully reliable event source, so a poll has to stay as a backstop either way. The reasoning is recorded on `POLL_INTERVAL`.

To confirm the new test guards the real defect, I restored the old offset-tailing loop and re-ran it. `detects_event_written_over_already_read_bytes` failed with "event written over already-read bytes was missed", while `detects_appended_event` still passed.

This narrows the race but does not close it. Replicas of one service still share one file, and so do several SDK-linked processes in one container. Only a per-process output path closes it. This PR therefore does not auto-close #216, which stays open for that work.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, BTreeSet};
+use std::io::Read;
 use std::path::{Path, PathBuf};
 
 use color_eyre::Section;
@@ -299,18 +300,14 @@ async fn validate_compose(
     // process tree — so --keep-running leaves a working project behind.
     up_child.kill_group().await.ok();
 
-    let test_result = match result {
-        Ok(SetupOutcome::Found) => {
-            eprintln!("Setup-complete event detected.");
-            // Discover test commands only now: setup-complete means the
-            // project came up, so every container exists and discovery reads
-            // settled filesystems instead of racing container startup.
-            discover_scripts(rt, &compose, overlay, temp_dir)
-                .and_then(|scripts| validate_test_scripts(&scripts))
-        }
-        Ok(SetupOutcome::TimedOut) => Err(eyre!("timed out waiting for setup-complete event")),
-        Err(e) => Err(e),
-    };
+    let test_result = result.and_then(|()| {
+        eprintln!("Setup-complete event detected.");
+        // Discover test commands only now: setup-complete means the project
+        // came up, so every container exists and discovery reads settled
+        // filesystems instead of racing container startup.
+        discover_scripts(rt, &compose, overlay, temp_dir)
+            .and_then(|scripts| validate_test_scripts(&scripts))
+    });
 
     if test_result.is_ok() {
         eprintln!("Setup validation successful.");
@@ -830,23 +827,23 @@ fn find_jsonl_files(root: &Path) -> Result<Vec<PathBuf>> {
 /// closes that window; see [`watch_for_setup_complete`].
 const POLL_INTERVAL: Duration = Duration::from_millis(25);
 
-/// Why [`watch_for_setup_complete`] stopped watching.
-#[derive(Debug, PartialEq, Eq)]
-enum SetupOutcome {
-    /// The setup-complete event was seen.
-    Found,
-    /// The deadline passed without the event.
-    TimedOut,
-}
+/// How much of one output file to read.
+///
+/// Setup-complete arrives early in a run, so it is not expected to sit behind a
+/// megabyte of other events. Capping the read keeps a chatty — or corrupt —
+/// file from being pulled into memory on every poll. An event past the cap is
+/// not seen, and the watch times out as though it never arrived.
+const MAX_READ_BYTES: u64 = 1024 * 1024;
 
 /// Watch `.jsonl` files anywhere under the given directory for setup-complete.
+/// Returns once the event is seen, and errors when the deadline passes first.
 ///
-/// Reads every file in full on every poll, and remembers nothing between
-/// polls. The SDK's local-file handler opens its output without `O_APPEND` and
-/// truncates it during initialization, so every SDK-linked process writes from
-/// offset 0. Nothing a previous poll learned about a file's contents stays
-/// true, so tracking read offsets would only let the watcher skip a region a
-/// later writer had refilled. These files hold a handful of short lines until
+/// Reads every file on every poll, and remembers nothing between polls. The
+/// SDK's local-file handler opens its output without `O_APPEND` and truncates
+/// it during initialization, so every SDK-linked process writes from offset 0.
+/// Nothing a previous poll learned about a file's contents stays true, so
+/// tracking read offsets would only let the watcher skip a region a later
+/// writer had refilled. These files hold a handful of short lines until
 /// setup-complete arrives, which is the whole period this runs for, so reading
 /// them again costs nothing worth optimizing.
 ///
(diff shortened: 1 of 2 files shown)


src/validate.rs
```diff
@@ -301,15 +301,15 @@ async fn validate_compose(
...
             eprintln!("Setup-complete event detected.");
             // Discover test commands only now: setup-complete means the
             // project came up, so every container exists and discovery reads
             // settled filesystems instead of racing container startup.
             discover_scripts(rt, &compose, overlay, temp_dir)
                 .and_then(|scripts| validate_test_scripts(&scripts))
         }
-        Ok(false) => Err(eyre!("timed out waiting for setup-complete event")),
+        Ok(watch) => Err(setup_timeout_error(&watch.rewritten)),
         Err(e) => Err(e),
```
carlsverre: The result should be an enum that clearly encodes: Found, Timed out, Error

carlsverre (commented):
remove the "rewritten file" tracking part of this code. this is fragile and the error isn't something a customer should have to deal with. i'm going to investigate ways to improve this upstream. in the meantime I want this PR to just make snouty more reliably able to find setup complete events when it can.

src/validate.rs
carlsverre: Done — `watch_for_setup_complete` now returns `SetupOutcome::{Found, TimedOut}`, so the call site matches on `Found`, `TimedOut`, and `Err`. I kept the error as the `Result::Err` arm rather than a third enum variant, because the watcher uses `?` on the IO calls inside the loop. Say the word if you want the error folded into the enum instead.

carlsverre:
Pushed. The rewritten-file tracking is gone: `SetupWatch.rewritten`, `rewrote_earlier_bytes`, the SHA-256 digest, and `setup_timeout_error` are all removed, along with their tests. The timeout error is back to the plain `timed out waiting for setup-complete event`. What remains is only the reliability part you asked for: read each file from the start when its length or mtime moved, and poll every 25ms instead of every 100ms.

`watch_for_setup_complete` also returns a `SetupOutcome` enum now, per the inline comment.

Both @claude comments are moot as a result — the `indented_list` nit was about `setup_timeout_error`, and the SHA-256 note was about the digest. Neither exists anymore. The other three indent+join sites the nit mentions are pre-existing, so I left them alone.

Rebased onto main for the 0.7.0-rc.2 release and force-pushed, so the branch is two clean commits rather than a fix followed by a revert. The changelog entry is trimmed to match.

`cargo test`, `cargo clippy --all-targets`, and `cargo fmt --check` are clean on the rebased branch, including the 13 spec tests that drive real docker and podman validate runs.

src/validate.rs
```diff
@@ -841,49 +822,65 @@ fn find_jsonl_files(root: &Path) -> Result<Vec<PathBuf>> {
...
         if tokio::time::Instant::now() >= deadline {
-            return Ok(false);
+            return Ok(SetupOutcome::TimedOut);
         }
 
         for path in find_jsonl_files(output_dir)? {
-            let previous_offset = offsets.get(&path).copied().unwrap_or(0);
-            let Some(mut file) = open_jsonl_file(&path)? else {
-                continue;
+            let content = match std::fs::read(&path) {
```
carlsverre: Read up to a max of 1MB for safety. We don't expect a setup_complete event to appear after 1MB of events in general.

src/validate.rs
```diff
@@ -841,49 +822,65 @@ fn find_jsonl_files(root: &Path) -> Result<Vec<PathBuf>> {
...
+/// closes that window; see [`watch_for_setup_complete`].
+const POLL_INTERVAL: Duration = Duration::from_millis(25);
+
+/// Why [`watch_for_setup_complete`] stopped watching.
+#[derive(Debug, PartialEq, Eq)]
+enum SetupOutcome {
+    /// The setup-complete event was seen.
+    Found,
+    /// The deadline passed without the event.
+    TimedOut,
```
carlsverre: rename to "Timeout"

src/validate.rs
```diff
@@ -841,49 +822,65 @@ fn find_jsonl_files(root: &Path) -> Result<Vec<PathBuf>> {
...
     output_dir: &Path,
     deadline: tokio::time::Instant,
-) -> Result<bool> {
-    let mut offsets = BTreeMap::new();
-
+) -> Result<SetupOutcome> {
     loop {
         if tokio::time::Instant::now() >= deadline {
-            return Ok(false);
+            return Ok(SetupOutcome::TimedOut);
```
carlsverre: would this be simpler if you just return an eyre error "timed out"? you only use it to emit a custom error message anyways. why not do that here and then you can kill the enum and simplify validate_compose?

src/validate.rs
carlsverre: Done — `MAX_READ_BYTES` is 1MiB, applied with `file.take(MAX_READ_BYTES)`. A line split by the cap fails to parse and is ignored, same as any partial trailing line. `ignores_an_event_past_the_read_cap` pins that an event behind the cap is deliberately not found.

src/validate.rs
carlsverre: Superseded by your next comment — the enum is gone, so there is no variant left to rename. The timeout is now `bail!("timed out waiting for setup-complete event")` inside the watcher.

src/validate.rs
carlsverre: Agreed, and done. `watch_for_setup_complete` returns `Result<()>` and bails on the deadline. The enum is deleted, and `validate_compose` collapses from a three-arm match to `result.and_then(...)` on the success path.

PR #221, opened 2026-08-14
Add snouty runs wait

### LINKED ISSUE #203: snouty wait
I've seen claude write a million polling loops to wait for a run to complete. this may be partially moot with streaming search, but either way having some kind of simple "wait for run to complete" command which internalizes the optimal way to do this would be way better for agents.

we should also evaluate `snouty launch --block-until-complete`, although a separate command makes it easier to resume after interruption (which happens frequently with long running agent sessions)

carlsverre:
This adds `snouty runs wait RUN_ID`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete), then print the run's details — the same output as `runs show`, and the same JSON under `--json`. The command exits 0 for every terminal state; the run's outcome is in the output, not the exit code. Progress goes to stderr: one line per status change, plus a one-time warning when a run has been active for more than twice its `antithesis.duration`. The wait holds no state beyond the run id, so an interrupted wait resumes with a re-run of the same command. `snouty launch --json | jq -r .runId | xargs snouty runs wait --json` is now a complete launch-to-report pipeline.

The poll bypasses snouty's response cache. The live API serves run detail with `Cache-Control: private, max-age=3600`, so a wait that polled through the cache could re-read a stale status for up to an hour. A new `AntithesisApi::new_uncached` constructor disables the cache for this command, and `build`'s cache parameter becomes an explicit `ResponseCache::{Default, Disabled, Dir}` enum — its old `Option<PathBuf>` doc claimed `None` disables caching, but `None` actually enabled the XDG default.

Decisions from the issue discussion:

- A run that reports status `unknown` fails the command and names the status the server sent; the caller decides what to do. A novel status string also fails, because the closed `RunStatus` enum rejects it at parse time.
- `--poll-interval` defaults to 60 seconds with a 30-second floor, enforced by clap.
- The wait is unbounded unless `--timeout SECONDS` is given. The final poll lands at the deadline instead of up to a full interval later, and the timeout error names the last status and the resume command.
- No `launch --block-until-complete` flag: a separate command resumes after interruption, which a launch flag cannot.

Tests split per the issue discussion: `specs/wait.txt` pins the static behavior against the static mock (terminal states, `unknown`, run-not-found, `--timeout 0`, help text, argument validation), and Rust integration tests in `src/runs.rs` transition a wiremock server between polls (starting → in_progress → completed, `unknown` mid-wait, timeout, 500 pass-through) at millisecond intervals. A new `run-unknown-status` mock fixture backs the spec. `RunStatus` gains an `is_terminal` helper. The gallery gains `runs-wait` and `help-runs-wait` stories. The staging spec suite passes against a live tenant, and a manual smoke test on a real completed run prints the detail card and exits 0.

fixes #203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Add `snouty runs wait`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete), then print its details like `runs show`. The wait is unbounded unless `--timeout` is given, `--poll-interval` tunes the check frequency (30-second floor), a run that reports status `unknown` fails the command, and a run active for more than twice its scheduled duration gets a one-time warning ([#221](https://github.com/antithesishq/snouty/pull/221))
 - `snouty validate` reads the SDK output file from the start instead of tailing it, so a setup-complete event written over bytes snouty had already read is still detected ([#218](https://github.com/antithesishq/snouty/pull/218))
 - Add `snouty runs exec`: run a bash script in a run's live session at a given moment, on a fresh branch of the multiverse. The command is gated behind the `runs-exec` unstable feature (`SNOUTY_UNSTABLE_FEATURES=runs-exec`), because the API it calls is unstable and unavailable on most tenants ([#208](https://github.com/antithesishq/snouty/pull/208))
 - snouty now requires Docker Compose v2.24.7 or newer, checked up front ([#214](https://github.com/antithesishq/snouty/pull/214))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Add `snouty runs wait`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete), then print its details like `runs show`. The wait is unbounded unless `--timeout` is given, `--poll-interval` tunes the check frequency (30-second floor), a run that reports status `unknown` fails the command, and a run active for more than twice its scheduled duration gets a one-time warning ([#221](https://github.com/antithesishq/snouty/pull/221))
+- Add `snouty runs wait`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete) ([#221](https://github.com/antithesishq/snouty/pull/221))
 - `snouty validate` reads the SDK output file from the start instead of tailing it, so a setup-complete event written over bytes snouty had already read is still detected ([#218](https://github.com/antithesishq/snouty/pull/218))
 - Add `snouty runs exec`: run a bash script in a run's live session at a given moment, on a fresh branch of the multiverse. The command is gated behind the `runs-exec` unstable feature (`SNOUTY_UNSTABLE_FEATURES=runs-exec`), because the API it calls is unstable and unavailable on most tenants ([#208](https://github.com/antithesishq/snouty/pull/208))
 - snouty now requires Docker Compose v2.24.7 or newer, checked up front ([#214](https://github.com/antithesishq/snouty/pull/214))
--- a/src/api.rs
+++ b/src/api.rs
@@ -30,7 +30,7 @@ mod generated {
 pub(crate) use generated::types::Params as RunParams;
 pub use generated::types::{
     BuildLogLine, Event, EventProperty, Moment, NonEventProperty, Property, PropertyStatus,
-    RunDetail, RunSummary,
+    RunDetail, RunStatus, RunSummary,
 };
 pub use progenitor_client::ByteStream;
 
@@ -67,95 +67,15 @@ pub enum VersionError {
     Unreachable(String),
 }
 
-/// A run's lifecycle status.
-///
-/// The documented statuses get variants; any other value the server sends is
-/// preserved verbatim as [`RunStatus::Other`], so a status this snouty
-/// predates never fails a whole response parse. Wired into the generated
-/// client by the `run_status` conversion in build.rs, like [`VTime`].
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum RunStatus {
-    Starting,
-    InProgress,
-    Completed,
-    Cancelled,
-    Incomplete,
-    Unknown,
-    /// A status this snouty does not know, carried verbatim.
-    Other(String),
-}
-
 impl RunStatus {
-    /// The wire string, e.g. `in_progress`.
-    pub fn as_str(&self) -> &str {
-        match self {
-            RunStatus::Starting => "starting",
-            RunStatus::InProgress => "in_progress",
-            RunStatus::Completed => "completed",
-            RunStatus::Cancelled => "cancelled",
-            RunStatus::Incomplete => "incomplete",
-            RunStatus::Unknown => "unknown",
-            RunStatus::Other(raw) => raw,
-        }
-    }
-
     /// Whether the run has stopped and its status will not change again.
     /// `unknown` is not terminal: it is the server saying it cannot classify
-    /// the run. Neither is [`RunStatus::Other`], which snouty cannot judge.
-    pub(crate) fn is_terminal(&self) -> bool {
+    /// the run.
+    pub(crate) fn is_terminal(self) -> bool {
         matches!(
             self,
             RunStatus::Completed | RunStatus::Cancelled | RunStatus::Incomplete
         )
-    }
-}
-
-impl std::fmt::Display for RunStatus {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(self.as_str())
-    }
-}
-
-impl std::str::FromStr for RunStatus {
-    type Err = std::convert::Infallible;
-
-    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
-        Ok(match s {
-            "starting" => RunStatus::Starting,
-            "in_progress" => RunStatus::InProgress,
-            "completed" => RunStatus::Completed,
-            "cancelled" => RunStatus::Cancelled,
-            "incomplete" => RunStatus::Incomplete,
-            "unknown" => RunStatus::Unknown,
-            other => RunStatus::Other(other.to_string()),
-        })
-    }
-}
-
-impl From<RunStatus> for String {
-    fn from(status: RunStatus) -> Self {
-        match status {
-            RunStatus::Other(raw) => raw,
-            known => known.as_str().to_owned(),
-        }
-    }
-}
-
-impl serde::Serialize for RunStatus {
-    fn serialize<S: serde::Serializer>(
-        &self,
-        serializer: S,
-    ) -> std::result::Result<S::Ok, S::Error> {
-        serializer.serialize_str(self.as_str())
-    }
-}
-
-impl<'de> serde::Deserialize<'de> for RunStatus {
-    fn deserialize<D: serde::Deserializer<'de>>(
-        deserializer: D,
-    ) -> std::result::Result<Self, D::Error> {
-        let raw = String::deserialize(deserializer)?;
-        Ok(raw.parse().expect("RunStatus::from_str is infallible"))
     }
 }
 
@@ -582,7 +502,7 @@ impl AntithesisApi {
             request = request.after(cursor);
         }
         if let Some(ref status) = opts.status {
-            request = request.status(status.clone());
+            request = request.status(*status);
         }
         if let Some(ref launcher) = opts.launcher {
             request = request.launcher(launcher.clone());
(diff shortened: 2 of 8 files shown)


src/api.rs
```diff
@@ -185,6 +198,23 @@ fn normalize_property(property: Property) -> Result<Property> {
...
+    /// Caching is disabled when `XDG_RUNTIME_DIR` is unusable.
+    Default,
+    /// No response cache: every request hits the server. For poll loops that
+    /// must observe fresh state — the server marks run detail cacheable
+    /// (`Cache-Control: private, max-age=3600`), so a cached poll could
+    /// re-read the same status for up to an hour.
+    Disabled,
+    /// Cache under this directory. Only tests construct this variant (the
+    /// commands all use `Default` or `Disabled`), hence the dead-code allow.
+    #[allow(dead_code)]
```
carlsverre: can we make this cfg[test]?

src/cli.rs
```diff
@@ -563,6 +564,42 @@ Incomplete runs also show the failure moment (Failure Hash/VTime) to pass to
...
     },
 
+    /// Wait for a run to reach a terminal state
+    #[command(
+        long_about = r#"Wait for a run to reach a terminal state (completed, cancelled, or incomplete).
+
+Polls the run's status until it is terminal, then prints the run's details —
+the same output as `runs show` — and exits 0 whatever the terminal state is;
+the run's outcome is in the output, not the exit code. A run that reports
+status `unknown` fails the command instead: snouty cannot tell whether such a
```
carlsverre: I disagree that this should then print the output of runs show.

this command should stand alone and simply report the final status if terminal or an error if unknown/some other error occurs

src/cli.rs
```diff
@@ -563,6 +564,42 @@ Incomplete runs also show the failure moment (Failure Hash/VTime) to pass to
...
+    )]
+    Wait {
+        /// Run ID
+        run_id: String,
+
+        /// Seconds between status checks
+        // The 30-second floor keeps a scripted wait from hammering the API:
+        // polling faster cannot observe a run (which takes minutes to hours)
+        // any sooner.
+        #[arg(long, default_value_t = 60, value_parser = clap::value_parser!(u64).range(30..))]
```
carlsverre: 1. refactor ReportDuration into HumanDuration in another PR, then rebase this PR on top of it
2. use HumanDuration to parse this field
3. set the minimum and default to 1 minute

src/cli.rs
```diff
@@ -563,6 +564,42 @@ Incomplete runs also show the failure moment (Failure Hash/VTime) to pass to
...
+        poll_interval: u64,
+
+        /// Maximum seconds to wait before the command fails; without it the
+        /// wait is unbounded
+        #[arg(long)]
+        timeout: Option<u64>,
```
carlsverre: use HumanDuration here also

src/runs.rs
```diff
@@ -364,6 +379,117 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+
+        if run.status == RunStatus::Unknown {
+            return Err(user_error(format!(
+                "run {run_id} reported status \"{}\"",
+                status_label(run.status)
+            ))
+            .note("snouty cannot tell whether this run will still make progress")
+            .suggestion(format!(
+                "inspect the run with `snouty runs show {run_id}`, or resume waiting with \
+                 `snouty runs wait {run_id}`"
```
carlsverre: resume waiting is almost certainly wrong as it will likely just loop back to this point. the user should inspect the run

src/runs.rs
```diff
@@ -364,6 +379,117 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+
+        if last_status != Some(run.status) {
+            eprintln!("run {run_id} is {}", status_label(run.status));
+            last_status = Some(run.status);
+        }
+
+        if run.status == RunStatus::Unknown {
+            return Err(user_error(format!(
+                "run {run_id} reported status \"{}\"",
+                status_label(run.status)
```
carlsverre: delete the status_label function from the codebase, RunStatus impl's display

carlsverre: is it possible to get serde to preserve the raw status rather than casting it to Unknown?

src/runs.rs
```diff
@@ -364,6 +379,117 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+        }
+
+        let now = tokio::time::Instant::now();
+        match deadline {
+            Some(deadline) if now >= deadline => {
+                return Err(user_error(format!(
+                    "timed out waiting for run {run_id}; it is still {}",
+                    status_label(run.status)
+                ))
+                .suggestion(format!("resume waiting with `snouty runs wait {run_id}`")));
```
carlsverre: remove this suggestion

src/runs.rs
```diff
@@ -364,6 +379,117 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+            }
+            // Clip the final sleep so the last poll lands at the deadline,
+            // not up to a full interval later.
+            Some(deadline) => tokio::time::sleep(poll_interval.min(deadline - now)).await,
+            None => tokio::time::sleep(poll_interval).await,
```
carlsverre: this optimization is too much, just use a tokyo Ticker for the poll interval

src/runs.rs
```diff
@@ -364,6 +379,117 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+    run_id: &str,
+    poll_interval: std::time::Duration,
+    timeout: Option<std::time::Duration>,
+) -> Result<RunDetail> {
+    let deadline = timeout.map(|t| tokio::time::Instant::now() + t);
+    let mut last_status = None;
+    let mut warned_over_schedule = false;
+
+    loop {
+        let run = match api.get_run(run_id).await {
```
carlsverre, line 441: A hung network request can cause this cmd to deadlock

I think you might be able to wrap wait_for_run with tokio::time::timeout in the parent function allowing wait_for_run to just loop forever.

Also consider restructuring wait_for_run into some kind of tokio::time::interval based loop. see if that's cleaner

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Add `snouty runs wait`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete), then print its details like `runs show`. The wait is unbounded unless `--timeout` is given, `--poll-interval` tunes the check frequency (30-second floor), a run that reports status `unknown` fails the command, and a run active for more than twice its scheduled duration gets a one-time warning ([#221](https://github.com/antithesishq/snouty/pull/221))
```
carlsverre:
```suggestion
- Add `snouty runs wait`: poll a run until it reaches a terminal state (completed, cancelled, or incomplete) ([#221](https://github.com/antithesishq/snouty/pull/221))
```

src/api.rs
carlsverre: Done: the variant and its match arm in `build()` are `#[cfg(test)]` now.

src/cli.rs
carlsverre: Done. Human output is one line — `run <id> is <status>` — and `--json` emits `{"run_id", "status"}`. Status-change heartbeats stay on stderr.

src/cli.rs
carlsverre: Done. #227 renames `ReportDuration` to `HumanDuration`, and this PR is rebased on it (its base is the `human-duration` branch until #227 merges). `--poll-interval` parses as `HumanDuration`, and its default and floor are both 1 minute.

src/cli.rs
carlsverre: Done: `--timeout` is a `HumanDuration` too (e.g. `--timeout 2h`; a bare number is minutes, like `--duration`).

src/runs.rs
carlsverre: Done: the suggestion is now only `snouty runs show <id>`.

src/runs.rs
carlsverre: Done: `status_label` is deleted; call sites use `RunStatus`'s `Display` directly.

carlsverre: It was actually worse than casting: the generated enum was closed, so `unknown` only matched the literal string `"unknown"`, and any other value failed the whole response parse — a new server-side status would have broken `runs list` entirely.

Fixed by taking `RunStatus` out of the generated code. `build.rs` inlines the `Run_Status` schema and maps it onto a handwritten `crate::api::RunStatus` via `with_conversion` — the same mechanism as `VTime`. The enum is open: an unrecognized value deserializes as `RunStatus::Other(String)` and round-trips verbatim, so `runs wait` now reports exactly what the server sent (`run <id> reported status "paused"`). `--status` still rejects anything that isn't a documented value, and a build.rs pin fails the build if a spec refresh changes the status list.

src/runs.rs
carlsverre: Done: the timeout error carries no suggestion.

src/runs.rs
carlsverre: Done, both suggestions taken: `--timeout` wraps the whole wait in `tokio::time::timeout` — covering any in-flight request, since the client deliberately has no read timeout — and `wait_for_run` is a `tokio::time::interval` loop (`MissedTickBehavior::Delay`) that never returns on its own. A test pins the hung-request case with a server that answers after an hour.

One open question from smoke testing: a live poll hit a transient 500, which fails the wait immediately (same as every other command). For a multi-hour wait, one blip killing the command reintroduces the interruption pain this command exists to remove. Should `wait` retry transport errors and 5xx until the timeout, failing fast only on 404 and unknown statuses? Happy to do that as a follow-up.

CHANGELOG.md
carlsverre: Applied verbatim.

src/runs.rs
carlsverre: Reverted in a910d96 per our discussion: `unknown` is a documented value of the spec's `Run_Status` enum, so there is no raw server string to preserve. The generated closed enum is back, and `wait` fails on `unknown` as before.

src/runs.rs
```diff
@@ -364,6 +362,128 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
     Ok(())
 }
 
+/// `runs wait`: poll the run until it reaches a terminal state, then report
+/// the final status. The client is uncached (see
+/// [`crate::api::ResponseCache::Disabled`]) so every poll observes fresh state.
+async fn cmd_runs_wait(
+    run_id: &str,
+    poll_interval: HumanDuration,
+    timeout: Option<HumanDuration>,
```
carlsverre: take duration at this point, we don't need humanduration anymore

src/runs.rs
```diff
@@ -364,6 +362,128 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+        outln!("run {} is {}", run.run_id, run.status)?;
+    }
+
+    Ok(())
+}
+
+/// [`wait_for_run`] bounded by `timeout`. The timeout wraps the whole wait —
+/// including any in-flight request, since the HTTP client has no read timeout
+/// (see `CONNECT_TIMEOUT` in api.rs) — so a hung request cannot outlive it.
+async fn wait_with_timeout(
```
carlsverre: inline this function into cmd_runs_wait

src/runs.rs
```diff
@@ -5598,13 +5718,13 @@ mod tests {
...
 
     #[test]
-    fn status_label_covers_every_variant() {
-        assert_eq!(status_label(RunStatus::Completed), "completed");
-        assert_eq!(status_label(RunStatus::Incomplete), "incomplete");
-        assert_eq!(status_label(RunStatus::InProgress), "in_progress");
-        assert_eq!(status_label(RunStatus::Starting), "starting");
-        assert_eq!(status_label(RunStatus::Cancelled), "cancelled");
-        assert_eq!(status_label(RunStatus::Unknown), "unknown");
+    fn run_status_display_covers_every_variant() {
```
carlsverre: delete this test

src/runs.rs
```diff
@@ -364,6 +362,128 @@ fn open_run_report(run: &RunDetail, json: bool) -> Result<()> {
...
+    // A poll delayed by a slow response schedules the next one a full
+    // interval later, instead of firing early to catch up.
+    poll.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+    let mut last_status = None;
+    let mut warned_over_schedule = false;
+
+    loop {
+        poll.tick().await;
+
+        let run = match api.get_run(run_id).await {
```
carlsverre, line 441: File an issue to handle transient network errors by integrating reqwest-retry into the API client for GET requests. 500s sadly should still fail.

src/runs.rs
carlsverre: Done in 89ba67d: `cmd_runs_wait` and `wait_for_run` take `std::time::Duration`; only `--timeout` stays `HumanDuration`, for its "timed out ... after 2h" error text.

src/runs.rs
carlsverre: Done: inlined into `cmd_runs_wait`. The two timeout tests now wrap `wait_for_run` with `tokio::time::timeout` directly.

src/runs.rs
carlsverre: Deleted.

src/runs.rs
carlsverre: Filed #230: reqwest-retry in the API client's middleware chain, GET requests only, transport errors only — HTTP statuses (500 included) still fail.

PR #222, opened 2026-08-14
Converge logs/events/search on one event-aware renderer

carlsverre:
Converge the human output of `runs logs`, `runs events`, and `runs search` on one event-aware renderer (`src/event_render.rs`). The renderer plugs into the shared `print_event_stream` pipeline from #206 and consumes the parsed `JsonStream` lines directly. `--json` output is unchanged and remains the scripting contract; the human formats change deliberately.

The renderer classifies each stream entry and renders it as a concise block, git-log style. A yellow `moment HASH VTIME` divider (full precision, pasteable into `runs logs`/`runs exec`/`snouty debug`) opens each timeline segment. Each event renders as one `VTIME [source] payload` line, with dim indented detail lines where a kind has more to say. Blank lines separate segments. Colors go through `console::style`, so pipes and `NO_COLOR` get plain text.

The classifier keys on payload fields and `source.name`, derived from ~135k raw events sampled across all five launchers on the staging tenant (compose and k8s):

- SDK assertions: a verdict badge per family. `hit:false` is a dim `CATALOG` registration; `FAIL` (red) only on a failing always/AlwaysOrUnreachable or a hit unreachable; `HIT` otherwise (green for a passing always, plain for sometimes/reachable).
- SDK guidance: `hit:false` is a dim `CATALOG GUIDANCE` line. An observation renders the drive on the value: `20 ↗ 1000` is the explorer pushing the value up toward its bound. The source inequality is deliberately not reconstructed — its strictness is unrecoverable and its direction depends on an assertion event a mid-stream reader may never see.
- Faults: type/name plus the details that matter per family (disruption, asymmetry, clock offset, latency, nodes, max duration); `restore` renders green.
- Container lifecycle: create/start/died with exit codes; the image (repository only) on create; verdict-less `health_status` probes render dim.
- Test composer: task started/finished with exit and runtime; other chatter as dim key=value.
- SDK handshake, setup-complete, fault-injector status; plain log text; unknown shapes fall back to their payload JSON, and rows reshaped by `map`/`narrow`/`fold` print as bare JSON.

Shared conventions live in one place now: vtime normalization and the fixed-width vtime cell (a character budget on Display itself — 8 columns by default, 18 under `--detail` — truncated, never rounded, so a value copied off the screen and pasted back as `--begin-vtime` lands on, never past, the line it came from; exponent-form values print whole because cutting their digits would drop the exponent and multiply the value), control-byte escaping with SUT SGR color sequences preserved (reset-terminated, colors-on only), and source labels (container or name, stream dropped; labels past 40 columns truncate with the full value on a dim detail line). `runs build-logs` renders through the same module.

Two flag changes:

- `-d/--detail` on logs/events/search: a full-width vtime on every line (18 columns holds every moment a real run reports whole), source locations, attached details JSON, the composer's captured stdout/stderr, and composer chatter expanded one key=value per line. Mutually exclusive with `--json` and `--raw`.
- `--raw` (now also on events/search) requires `--json` and passes the server's stream through without normalization — parsed records round-trip verbatim (serde_json's `preserve_order` + `float_roundtrip` keep the round-trip faithful; a vtime stays the server's string). The legacy raw renderer is deleted, and the search limit is still enforced client-side on the passthrough.

Specs cover the new formats, and the staging spec suite passes against tenant release 60.1.


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
...
...ware renderer. Each event is auto-classified (SDK assertions, guidance, SDK handshake, setup-complete, faults, container lifecycle, test-composer chatter, plain log text, unknown JSON) and rendered as a concise colored block: a `moment HASH VTIME` divider (full precision, pasteable into `runs logs`/`runs exec`/`snouty debug`) opens each timeline segment, and each event renders as one `VTIME [source] payload` line. Assertions carry a verdict badge — HIT/FAIL on evaluations, FAIL only on a failing always/unreachable, CATALOG on the SDK's startup registrations. SUT color codes survive in log text (restricted to SGR so they cannot corrupt the terminal). A new `-d/--detail` flag on all three commands adds full-precision vtimes, source locations, attached details JSON, and the composer's captured stdout/stderr. Source labels past 40 columns (generated k8s pod names) truncate with the full value on a dim detail line. `runs build-logs` shares the same visual grammar. `--raw` (now also on `runs events` and `runs search`) requires `--json` and passes the server's NDJSON through verbatim. Scripts should use `--json`, which is unchanged ([#222](https://github.com/antithesishq/snouty/pull/222))
```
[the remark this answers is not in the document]
carlsverre: Stale — the review ran against the pre-link head. Commit 9329f19 appends the ([#222]) link to the entry.

src/event_render.rs
```diff
@@ -0,0 +1,1792 @@
...
+            .map(|json| style(format!("details {json}")).dim().to_string())
+            .into_iter()
+            .collect()
+    } else {
+        Vec::new()
+    };
+    RenderedPayload {
+        headline: headline.trim_end().to_string(),
+        details,
+    }
```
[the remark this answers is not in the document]
carlsverre: Fixed in 6863743. `details_line`/`details_detail_line` now hold the shared rendering; assert/setup/fault all call them.

src/event_render.rs
```diff
@@ -0,0 +1,1792 @@
...
+        // The truncated source's full value leads the detail lines, right
+        // under the cell it overflowed.
+        let overflow = source
+            .overflow
+            .map(|overflow| style(overflow).dim().to_string());
+        for detail in overflow.into_iter().chain(rendered.details) {
+            out.push('\n');
+            out.push_str(&format!("{:DETAIL_INDENT$}{detail}", ""));
+        }
+        self.wrote_block = true;
```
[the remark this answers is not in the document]
carlsverre: Fixed in 6863743. `full_vtime` is computed once per entry; the divider and the cell derive from it, and `format_vtime_cell` is gone.

src/event_render.rs
```diff
@@ -0,0 +1,1792 @@
...
+            details.push(style(format!("@ {location}")).dim().to_string());
+        }
+        if let Some(json) = &summary.details {
+            details.push(style(format!("details {json}")).dim().to_string());
+        }
+    }
+    RenderedPayload { headline, details }
+}
+
+/// Guidance rendering follows ~/guidance-rendering.md: `hit: false` is the
```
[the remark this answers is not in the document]
carlsverre: Fixed in f49203d. The doc comment describes the rules inline and points at `render_guidance_expression` instead of the local note.

src/runs/event_search.rs
```diff
@@ -23,15 +23,29 @@ use futures_util::{StreamExt, TryStreamExt};
...
+pub(super) struct EventOutput {
+    pub json: bool,
+    /// Pass the server's stream through without normalization (parsed
+    /// records round-trip verbatim). Requires `json`; the commands refuse
+    /// the combination up front.
+    pub raw: bool,
+    /// Detailed human rendering: full-precision vtime on every line, source
+    /// locations, and attached details JSON.
+    pub detail: bool,
+}
```
carlsverre: seems like this should be an enum:
- Human { detail: bool }
- Json { raw: bool }

src/runs/event_search.rs
```diff
@@ -44,24 +58,42 @@ pub(super) async fn print_event_stream(
...
+        while let Some(value) = lines.try_next().await? {
+            seen += 1;
+            let rendered = if output.json {
+                value.to_string()
+            } else {
+                renderer.render_entry(&value)
+            };
+            writeln!(stdout, "{rendered}")?;
+            stdout.flush()?;
+        }
```
carlsverre: it would be nice to converge this renderer with the cmd_log renderer which is very similar. the only difference is fault annotation which you could add as a bool to `EventOutput::Json{annotate_faults: bool}`

src/runs.rs
```diff
@@ -1403,8 +1387,8 @@ async fn cmd_runs_build_logs(
                 wrote_any = true;
                 let ts = format_local_str(entry["timestamp"].as_str().unwrap_or(""));
                 let stream = entry["stream"].as_str().unwrap_or("out");
-                let text = sanitize(entry["text"].as_str().unwrap_or(""));
-                writeln!(stdout, "{ts} [{stream}] {text}")?;
+                let text = entry["text"].as_str().unwrap_or("");
+                writeln!(stdout, "{}", render_build_log_line(&ts, stream, text))?;
```
carlsverre: Can we make this an event renderer so we can use print_event_stream?

src/runs.rs
```diff
@@ -1472,6 +1464,8 @@ async fn cmd_runs_events(
     limit: Option<NonZeroU64>,
     settings: &Settings,
     OutputOptions { json, verbose }: OutputOptions,
+    raw: bool,
+    detail: bool,
```
carlsverre: use the typesystem properly.

we have raw, json, and detail.

raw can only be specified with json
detail can only be specified with !json

design a proper enum that encodes this concept and use it everywhere

AGENTS.md
```diff
@@ -124,14 +124,6 @@ uv sync                             # so ./.venv holds the dependencies
...
 ```
 
-## AI Coding Workflow
-
-1. Follow the Rust coding conventions in this file at all times
-2. Ensure that all user-visible changes are covered by a CLI spec or integration test
-3. Update gen-gallery.py when changing CLI behavior
-4. Verify all tests catch what you want them to catch via targeted mutation testing
-5. Test, check, and format code before finishing
-
```
carlsverre, line 134: revert this deletion

CHANGELOG.md
carlsverre: change to:

- Converge the human output of `runs logs`, `runs events`, and `runs search` on one event-aware renderer. Each event is rendered as a concise colored block. A new `-d/--detail` flag on all three commands adds full-precision vtimes, source locations, attached details JSON, and the composer's captured stdout/stderr. `runs build-logs` shares the same visual grammar. ([#222](https://github.com/antithesishq/snouty/pull/222))

src/event_render.rs
```diff
@@ -0,0 +1,1783 @@
...
+/// Stateful renderer for one event stream. Feed it entries (and raw lines) in
+/// stream order; it returns the exact text to print for each, dividers and
+/// blank-line separation included.
+pub(crate) struct EventStreamRenderer {
+    detail: bool,
+    last_input_hash: Option<String>,
+    wrote_block: bool,
+}
+
+impl EventStreamRenderer {
```
carlsverre, line 405: Model renderer as writing to a Writer rather than returning a string. This lets you pass the writer down to different event renders and eliminate a ton of intermediate allocations. Look at Serde serialization and the stdlib's Debug/Display formatters as inspiration on how to design this API.

carlsverre (commented):
break up event_render into smaller nested specialized modules that can be more easily reviewed and tested. things like an ansi module focused on ansi escapes.

the vtime type should handle rendering itself to a full precision or truncated string format

review this code for accidentally quadratic behavior. we should not be looping over strings/outputs/values/etc multiple times while rendering. this will slow things down excessively on long outputs.

figure out a trait for event types that brings together the concepts of classification and rendering. put each event type in it's own nested module under event_render (event_render/{event type}.rs).

PR #223, opened 2026-08-14
Fix the gallery harness, and restore the API-key default in login

### LINKED ISSUE #219: Gallery harness: login stories need a PTY; two validate stories time out; one stale doctor check
The satisfaction-review gallery (`scripts/gen-gallery.py`) has 11 stories that fail for harness reasons, not output quality. They block the review from judging login and validate output.

**1. All 8 `login-*` stories fail with `Error: Cannot prompt for value when not running in a TTY`.**

Since #212 moved the login prompts to inquire, plain stdin no longer drives them. Every login story exits 1 with no transcript and no persisted state, so the gallery judges nothing. Fix: run the login captures under a PTY (the repo already has PTY test infrastructure from #212) and feed the scripted answers through it.

**2. `validate-unrecognized-command` and `validate-non-executable-command` time out before discovery.**

Both stories print `Error: timed out waiting for setup-complete event` instead of reaching test-command discovery, so the intended output (discovery failure naming the offending command) never appears. This looks like fixture/container timing on a slow host. Fix: raise the fixture timeout for these two stories, or retry once before writing the story.

**3. `doctor-legacy-auth`'s check token is stale.**

The check greps for the word `legacy`, but the message now says "username/password authentication is deprecated" (#212). The output itself reads well — the automated FAIL is a false alarm. Fix: update the check to match the current wording.

Context: found during the 2026-08-14 satisfaction review (91 stories, orbitinghail release 60.1). The other 80 stories generate and judge correctly.

carlsverre:
Eleven of the gallery's 91 stories failed for harness reasons, so the
satisfaction review could not judge `snouty login` or two `snouty validate`
paths at all. Three separate causes, and a fourth found while fixing them.

The eight `login-*` stories exited 1 with "Cannot prompt for value when not
running in a TTY". `snouty login` prompts through `inquire`, which engages only
when stdin is a terminal, so piping answers at it took the non-interactive path
and died at the first prompt. Their scripted answers were stale too: menu index
numbers for what is now an arrow-driven menu.

Interactive commands are now their own kind of gallery story, beside the goal
and help ones. Nothing in the section is named for login — that is the only TTY
command today, not the last one. A story carries a dialogue of `(wait for,
send)` steps; `pexpect` types them at a real pseudo-terminal and `pyte` replays
the byte stream, so a frame is the screen exactly as it stood at one prompt. The
same stream is written beside the story as an asciinema recording. Frames split
at prompt boundaries rather than keypresses, and they earn their place on the
prompts that collapse: the credential menu is erased once chosen, so the closing
screen shows only the label that won. The tenant is contacted for real, because
the menu holds a single sign-on entry only where the tenant enables it, so the
menu step finds its row on screen instead of counting keypresses. Three
scenarios remain of the eight; `src/login.rs` already covered profile scoping,
an invalid tenant, and repairing broken settings against a scripted prompter,
and those tests now also cover Esc skipping credential storage and the settings
backup.

`validate-unrecognized-command` and `validate-non-executable-command` timed out.
This was a fixture bug rather than host timing: `validate` waits for the
setup-complete event first and discovers test commands only after it, and
neither fixture emitted that event, so both always timed out and never reached
the discovery failure they exist to show. Their header comments still claimed
discovery ran before the wait. Both now emit the event and then stay up, the way
the `valid` sample does.

`doctor-legacy-auth` asked for the word "legacy", which doctor stopped printing
when username/password became "deprecated" (#212), so a false FAIL was reported
on good output.

Writing the reuse-the-stored-values dialogue surfaced a regression, and it is
the only user-facing change here. `prompt_for_value` offers the stored tenant
and repository back, but `prompt_for_api_key` took no previous value: a user who
re-ran `snouty login` had to paste the key again, and hitting enter stored an
empty one. `inquire` has no default for a masked prompt, so the reuse rides on a
hint plus an empty answer. The hint is built from the stored key — its own first
characters, then a fixed run of stars, e.g. `sk-S********` — so it shows
whatever prefix the key carries and does not disclose the key's length. A key
that came from `ANTITHESIS_API_KEY` is not offered back, because keeping it
would copy an ambient secret into the credentials file.

A satisfaction review of the repaired gallery then found three more things:

- `doctor` answered an unset tenant with "set ANTITHESIS_TENANT or add it to a
  settings file", and missing credentials with "ask Antithesis support for an
  API key". Both are true and both are the slow route; `snouty login` sets
  tenant, repository, and credentials in one pass and was never named. Each of
  the three checks now says so in its own words, because all three fire at once
  on an unconfigured machine and the same sentence three times reads as noise.
- A story's shell block joined arguments on a bare space, so `--launcher Basic
  Test git` was shown unquoted — not the command that ran, and it fails if
  pasted. The command line now goes through `shlex.join`.
- `runs-events-multi-match` asserted that two needles return strictly fewer rows
  than one. `runs events` matches a single needle on the server, which returns
  at most `--limit` events, and filters the rest locally, so when the
  single-needle search is capped both searches report the cap while describing
  different sets. Snouty explains this in its own output; the check was failing
  good behaviour. Both stories now pass an explicit `--limit`, and the check
  reports that narrowing was not assertable rather than failing.

The review's rendering axis no longer measures a widest line against a
140-column threshold. A number says a line is long, not whether that hurts: a
wrapped table row is a defect, a wrapped signed URL is not. The axis now asks
what the wrap breaks.

`scripts/` gains its first two third-party dependencies, `pexpect` and `pyte`,
and a `[tool.pyright]` stanza so `uvx pyright` resolves them from `./.venv`.

All 78 stories the gallery generates today pass their automated checks.

fixes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- `snouty login` offers a stored API key back at the prompt, shown as a masked hint, so hitting enter keeps it instead of saving an empty key ([#219](https://github.com/antithesishq/snouty/issues/219))
 - `snouty validate` reads the SDK output file from the start instead of tailing it, so a setup-complete event written over bytes snouty had already read is still detected ([#218](https://github.com/antithesishq/snouty/pull/218))
 - Add `snouty runs exec`: run a bash script in a run's live session at a given moment, on a fresh branch of the multiverse. The command is gated behind the `runs-exec` unstable feature (`SNOUTY_UNSTABLE_FEATURES=runs-exec`), because the API it calls is unstable and unavailable on most tenants ([#208](https://github.com/antithesishq/snouty/pull/208))
 - snouty now requires Docker Compose v2.24.7 or newer, checked up front ([#214](https://github.com/antithesishq/snouty/pull/214))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # Unreleased
 
-- `snouty login` offers a stored API key back at the prompt, shown as a masked hint, so hitting enter keeps it instead of saving an empty key ([#219](https://github.com/antithesishq/snouty/issues/219))
 - `snouty validate` reads the SDK output file from the start instead of tailing it, so a setup-complete event written over bytes snouty had already read is still detected ([#218](https://github.com/antithesishq/snouty/pull/218))
 - Add `snouty runs exec`: run a bash script in a run's live session at a given moment, on a fresh branch of the multiverse. The command is gated behind the `runs-exec` unstable feature (`SNOUTY_UNSTABLE_FEATURES=runs-exec`), because the API it calls is unstable and unavailable on most tenants ([#208](https://github.com/antithesishq/snouty/pull/208))
 - snouty now requires Docker Compose v2.24.7 or newer, checked up front ([#214](https://github.com/antithesishq/snouty/pull/214))
--- a/src/login.rs
+++ b/src/login.rs
@@ -161,10 +161,15 @@ async fn do_cmd_login(
         )?,
     };
 
+    // Whatever credentials are already in reach, so the menu can default to the
+    // kind last used and the key prompt can offer the stored key back. The error
+    // is discarded on purpose: nothing reads it, and having none is the ordinary
+    // state on a first login rather than a failure.
     let current_credentials = AuthenticationInfo::for_ambient_configuration_with_attribution(
         profile_to_use.as_deref(),
         PasswordPolicy::Inspect,
-    );
+    )
+    .ok();
 
     // Capture the credential kind and where it was stored so the summary can name
     // both; `None` when the user chose to skip credential setup.
@@ -215,7 +220,7 @@ fn print_login_summary(
     profile: Option<&str>,
     settings_path: &Path,
     credentials: Option<AttributedValue<&str>>,
-    previous_credentials: Result<AttributedValue<AuthenticationInfo>>,
+    previous_credentials: Option<AttributedValue<AuthenticationInfo>>,
 ) {
     let scope = match profile {
         Some(p) => format!(" under profile `{p}`"),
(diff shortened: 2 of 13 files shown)


src/login.rs
```diff
@@ -366,9 +383,53 @@ async fn prompt_for_auth(
...
+/// How many stars stand in for the rest of a stored secret. A fixed count, so
+/// the hint does not disclose how long the secret is.
+const HINT_STARS: usize = 8;
+
+/// A display-only stand-in for a stored secret — its first few characters, then
+/// stars, e.g. `sk-F********`. The prompt shows this so the user can see that
+/// there is a stored value to keep, and recognize which one it is.
+fn secret_hint(secret: &str) -> String {
+    let visible: String = secret.chars().take(HINT_VISIBLE_CHARS).collect();
+    format!("{visible}{}", "*".repeat(HINT_STARS))
```
carlsverre: use the last 4 characters rather than the first 4. antithesis api keys always start with "antithesis_api_key_v2"

carlsverre: so the output should be something like ********9Pgw

src/login.rs
carlsverre, line 168: the fact that this stays as a result rather than being mapped to an option at this point I think is code smell. as far as I can tell the error is not used anywhere, so the error should be clearly discarded rather than propogated through the codebase. unless I'm wrong

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- `snouty login` offers a stored API key back at the prompt, shown as a masked hint, so hitting enter keeps it instead of saving an empty key ([#219](https://github.com/antithesishq/snouty/issues/219))
```
carlsverre: remove this changelog entry

src/login.rs
carlsverre: Done in ef888fb — the hint is now stars then the last four characters. Test keys throughout carry the real `antithesis_api_key_v2` prefix so they exercise the case that matters, instead of the `sk-…` placeholders that made a leading-character hint look like it worked.

src/login.rs
carlsverre: You are right, and I checked before changing it: all three consumers (`prompt_for_auth`, `stored_api_key`, `print_login_summary`) only branch on whether it is `Ok` — the error value is never read. Mapped to `Option` where it is produced, with a note on why it is dropped.

One thing I noticed but did not act on: `for_ambient_configuration_with_attribution` can fail for a real reason, not only "nothing configured" — e.g. `try_load_credentials_file()?` on an unparsable credentials file. Both before and after this change that is silently treated as "no credentials", and login then overwrites the file. Behaviour is unchanged here; flagging it in case it deserves its own fix.

CHANGELOG.md
carlsverre: Removed in ef888fb. Agreed — the API-key default is a regression against `snouty login`, which is itself unreleased, so there is nothing there for a reader of released notes.

PR #224, opened 2026-08-14
add nix package

figsoda:


.github/workflows/build.yml
carlsverre, line 328: can we add the nix build to the existing job? or does it need to be isolated for some reason?

package.nix
```diff
@@ -0,0 +1,55 @@
...
+  lib,
+  rustPlatform,
+  pkg-config,
+  openssl,
+  sqlite,
+}:
+
+rustPlatform.buildRustPackage {
+  pname = "snouty";
+  version = "${(lib.importTOML ./Cargo.toml).package.version}-unstable";
```
carlsverre, line 11: why -unstable? because it's based on main?
I suppose that makes sense... although i'd prefer to have the version track the snouty version and only show -unstable if it's not based on one of our release or pre-release tags

.github/workflows/build.yml
figsoda: i thought some parallelism would help speed, merging is totally fine and it will cache a little bit better

package.nix
figsoda: i don't think there's a way for nix to figure out whether this is tagged or unreleased

having `-unstable` or not i'm ok with either options, ur call

